PR #1, opened 2017-02-24
FarhadifarForce: pass through vertex element index for to obtain cell

wwaites:
And so, make it possible to have physical properties of the mesh depend on properties of the cell, e.g. type or other state


cell_based/src/population/forces/FarhadifarForce.cpp
```diff
@@ -196,18 +196,36 @@ double FarhadifarForce<DIM>::GetLineTensionParameter(Node<DIM>* pNodeA, Node<DIM
...
+}
+
 template<unsigned DIM>
 double FarhadifarForce<DIM>::GetAreaElasticityParameter()
 {
     return mAreaElasticityParameter;
 }
 
+template<unsigned DIM>
+double FarhadifarForce<DIM>::GetAreaElasticityParameter(unsigned elem_index, VertexBasedCellPopulation<DIM>& rVertexCellPopulation)
```
fcooper8472, line 212: This overload does not seem to be necessary: the new parameters are not used in any way.

cell_based/src/population/forces/FarhadifarForce.cpp
```diff
@@ -196,18 +196,36 @@ double FarhadifarForce<DIM>::GetLineTensionParameter(Node<DIM>* pNodeA, Node<DIM
...
+}
+
 template<unsigned DIM>
 double FarhadifarForce<DIM>::GetPerimeterContractilityParameter()
 {
     return mPerimeterContractilityParameter;
 }
 
+template<unsigned DIM>
+double FarhadifarForce<DIM>::GetPerimeterContractilityParameter(unsigned elem_index, VertexBasedCellPopulation<DIM>& rVertexCellPopulation)
```
fcooper8472, line 224: This overload does not seem to be necessary: the new parameters are not used in any way.

fcooper8472:
@wwaites apologies this has taken such a long time to get to.  We have recently moved development over to GitHub so I'm now going through the open PRs.

I assume this feature was never completed, because the new overloads do not use the parameters at all.  I'm going to close the pull request for now, but if this is still relevant, please feel free to comment/re-open.

wwaites:
I do not have a clear recollection of this -- it was a long time ago! Most likely this was a hook designed to be overridden in a subclass and to retain unchanged behaviour by default. I would have to indulge in some code archaeology to be sure.

PR #3, opened 2019-10-14
Added SteadyStateExtrapolationRunner

joeyshuttleworth:
This class finds the steady state behaviour of an action potential model. This is done by using linear regression on the state variables when it looks like we have exponential decay.

TestSteadyStateExtrapolationRunner.hpp provides an example which uses this class.


mirams:
Hi Joseph,

Thanks for working on this! We do pull request stuff (slightly confusingly) via a different site - https://chaste.cs.ox.ac.uk/trac/wiki/ChasteGuides/GitGuide the thing to do would be opening a ticket on our trac site and then putting changes onto a branch there. So I'll close this one...

PR #12, opened 2021-12-09
Ticket 3077

jmosborne:
This pulls in the work from the following tickets 
 #3077 #3078 #3079 #2087

Continuous testa and coverage pass https://chaste.cs.ox.ac.uk/buildbot/Nightly%20Coverage/coverage1573/


fcooper8472:
Hi @jmosborne, could you add a comment with some more context for this PR to help reviewing it?  Would be useful to have a:
- summary of the main changes
- what the use-case is for these changes

I'm also not quite clear what #3079 has to do with this PR?

jmosborne:
Sorry #3079 was in there by mistake.  This was work on 3 tickets adding small functionality we needed for a few projects. There were issues regenerating archives stoping me merging them one by one so I put them all in the same branch so we could use them together prior to merging to develop. 

Heres the functionality. 

#3077 Add Methods to remove ghost nodes 
#3078 Add bounded voroni Tessellation's on Cylindrical and Toroidal meshes.
#2087  Numerics for Cell Based:  Added variable (mUpdatingTimestepMultiple) which allows you to only remsh or have division/death on set timestep, also the ability to impose BCS mid timestep for use in higher order methods

Sometimes the ticket descriptions are poor as the wiki is often unusable in the day here so comments get forgotten. Hopefully not a problem going forward.

jmosborne:
Also a small fix to vertex remeshing from ticket #3080.

jmosborne:
Trying to bump this as will start to diverge die to other commits.

mesh/src/vertex/MutableVertexMesh.cpp
```diff
@@ -1276,23 +1276,170 @@ void MutableVertexMesh<ELEMENT_DIM, SPACE_DIM>::IdentifySwapType(Node<SPACE_DIM>
...
+                            p_end_node = this->mNodes[previous_previous_node_1];
+                        }
+                        else
+                        {
+                            /*
+                             * If this trips then the adjacent node isnt in both elements.
+                             * So check the other adjacent node by uncommenting the code blow and adding a new test
+                             * See #3080
+                             */
+                            NEVER_REACHED;
```
fcooper8472: What's happening below here?

mesh/src/vertex/Cylindrical2dVertexMesh.cpp
```diff
@@ -59,50 +58,192 @@ Cylindrical2dVertexMesh::Cylindrical2dVertexMesh(Cylindrical2dMesh& rMesh)
...
+        
+        // Loop over all nodes and check they're not outside [0,mWidth]
+        for (unsigned i=0; i<nodes.size(); i++)
+        {
+            CheckNodeLocation(nodes[i]);
+        }
+            
+        Cylindrical2dMesh extended_mesh(mpDelaunayMesh->GetWidth(0),nodes);
+
+// VtkMeshWriter<2, 2> mesh_writer2("TMP", "extended_delaunay_mesh", false);
```
fcooper8472: This looks like it shouldn't be here?

mesh/src/vertex/MutableVertexMesh.cpp
```diff
@@ -2096,20 +2241,43 @@ void MutableVertexMesh<ELEMENT_DIM, SPACE_DIM>::PerformT3Swap(Node<SPACE_DIM>* p
...
+    VertexElement<ELEMENT_DIM, SPACE_DIM>* p_element = this->GetElement(elementIndex);
+    unsigned num_nodes = p_element->GetNumNodes();
+
     // Store the index of the elements containing the intersecting node
     std::set<unsigned> elements_containing_intersecting_node = pNode->rGetContainingElementIndices();
 
     // Get the local index of the node in the intersected element after which the new node is to be added
     unsigned node_A_local_index = this->GetLocalIndexForElementEdgeClosestToPoint(pNode->rGetLocation(), elementIndex);
 
+// // Check to see if the node is connected to the containg element.
```
fcooper8472: This is also commented out

cell_based/test/simulation/TestNumericalMethods.hpp
```diff
@@ -64,8 +64,10 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "SmartPointers.hpp"
 #include "FileComparison.hpp"
 #include "PopulationTestingForce.hpp"
+#include "PlaneBoundaryCondition.hpp"
 #include "ForwardEulerNumericalMethod.hpp"
 #include "Warnings.hpp"
+#include "Debug.hpp"
```
fcooper8472: Debug header

cell_based/test/simulation/TestOffLatticeSimulation.hpp
```diff
@@ -295,6 +289,58 @@ class TestOffLatticeSimulation : public AbstractCellBasedWithTimingsTestSuite
         TS_ASSERT_EQUALS(SimulationTime::Instance()->GetTime(), 0.5);
     }
 
+
+  /**
```
fcooper8472: Indent

jmosborne:
Still passed Gcc Continuous build https://chaste.cs.ox.ac.uk/buildbot/builders/Continuous%20GCC/builds/656 after tidying

cell_based/src/population/MeshBasedCellPopulationWithGhostNodes.hpp
```diff
@@ -86,21 +96,12 @@ class MeshBasedCellPopulationWithGhostNodes : public MeshBasedCellPopulation<DIM
...
-public:
-
-    /**
-     * Set the ghost nodes by taking in a set of which nodes indices are ghost nodes.
-     *
-     * @param rGhostNodeIndices set of node indices corresponding to ghost nodes
-     */
-    void SetGhostNodes(const std::set<unsigned>& rGhostNodeIndices);
-
-private:
```
mirams, line 102: All the methods below here were private but are now public, is that intended?

fcooper8472 (commented):
@jmosborne I've had a look through this now.

My main query is how the numerical methods are being used.  My understanding is that the `AbstractNumericalMethod` class is essentially designed to say "am I doing forward Euler or Runge-Kutta".

But this PR adds methods to the `AbstractNumericalMethod` including `SetBoundaryConditions` and `SaveCurrentNodeLocations`.  To my mind these don't relate at all to the numerical method.

The boundary conditions are currently handled in `OffLatticeSimulation` where they are iterated over and the method `ImposeBoundaryCondition` is called, and that appears to take the old node locations and "corrects" them to conform to the boundary condition.

What I don't understand is why the numerical method would need to know about the boundary conditions?

Ditto `SaveCurrentNodeLocations` sounds as though it's something that does not belong in the numerical method.

jmosborne:
When your using a higher order method eg RK4 you do virtual moves where you update the node locations for a part timestep. In order to get the expected convergence you need to apply any boundary conditions on these virtual moves as well as the final move. If you don't you still only get order 1 convergence. See https://github.com/jmosborne/ChasteNumericalMethods/blob/master/src/RK4NumericalMethod.cpp
For an example of using the two methods you mention. 

I plan on moving the higher order methods to the core once the paper on them is submitted . But they are already covered by unit tests.

jmosborne:
From todays meeting we are now clear that the Numerical Methods should know about the Boundary conditions. This is to allow higher order methods to be able to apply the boundary conditions on substeps being taken. With this pull request these boundary conditions are cached in the numerical method every time solve() is called. This could be confusing and I have made a ticket https://chaste.cs.ox.ac.uk/trac/ticket/3090 to look at moving the Boundary conditions down into the numerical method. The reason im not doing it now is that it will need a refactor of how Special cell populations like the Buske update pop and the inmmered boundary method project work so makes sense to look at it once the IB work is further along.

mesh/src/vertex/MutableVertexMesh.cpp
```diff
@@ -1278,21 +1278,163 @@ void MutableVertexMesh<ELEMENT_DIM, SPACE_DIM>::IdentifySwapType(Node<SPACE_DIM>
...
+                        if (shared_elements.size()==2)
+                        {
+                            //This neighbouring node is in the same 2 elements so treat this as end node
+                            p_end_node = this->mNodes[previous_previous_node_1];
+                        }
+                        else
+                        {
+                            /*
+                             * If this trips then the adjacent node isn't in both elements.
+                             * So check the other adjacent node by uncommenting the code blow and adding a new test
```
fcooper8472: @jmosborne there is no code to uncomment below this line - could you check whether this comment makes sense and remove/alter it if not?

mesh/src/vertex/MutableVertexMesh.cpp
```diff
@@ -1278,21 +1278,163 @@ void MutableVertexMesh<ELEMENT_DIM, SPACE_DIM>::IdentifySwapType(Node<SPACE_DIM>
...
+                             * See #3080
+                             */
+                            NEVER_REACHED;
+                        }
+
+                        p_merged_node->rGetModifiableLocation() = p_end_node->rGetLocation();
+                        PerformNodeMerge(p_end_node,p_merged_node);  // This order as first node is kept and this has correct boundary information.
```
fcooper8472: @jmosborne I don't understand the comment on this line

jmosborne:
Sorted both these comments in latest revision. Comment was reffering to som interim code that was deleted during development.

fcooper8472 (approved):

PR #15, opened 2022-04-25
Get Chaste working on Ubuntu 22.04

fcooper8472:
These changes seem sufficient to get the continuous tests passing on Ubuntu 22.04 out-of-the-box.


jmpf:
All looks okay.  Would be interested to know why the MPI C++ change is necessary.

fcooper8472:
> All looks okay. Would be interested to know why the MPI C++ change is necessary.

Some changes to how MPI is found in CMake: on the latest Ubuntu it's not able to find the C API for some reason, but that doesn't matter to us.

Getting to the bottom of why not can wait for the major CMake overhaul that we'll have to do at some point.

PR #16, opened 2022-06-01
Migrate CI from Buildbot to GitHub actions

fcooper8472:
This is (very) unfinished: opening the PR to review changes to the GitHub actions workflows.


mirams (approved):
These exceptions look OK to me, I guess they slightly break the "users cause exceptions to throw", "developers hit asserts" and so have a tiny overhead for the check to happen perhaps https://chaste.cs.ox.ac.uk/trac/wiki/ExceptionStrategy but on the plus side it means that they get checked in production code too.

mesh/src/vertex/VertexMesh.cpp
```diff
@@ -1422,7 +1424,10 @@ unsigned VertexMesh<ELEMENT_DIM, SPACE_DIM>::GetLocalIndexForElementEdgeClosestT
 template <unsigned ELEMENT_DIM, unsigned SPACE_DIM>
 c_vector<double, 3> VertexMesh<ELEMENT_DIM, SPACE_DIM>::CalculateMomentsOfElement(unsigned index)
 {
-    assert(SPACE_DIM == 2); // LCOV_EXCL_LINE - code will be removed at compile time
+    if (SPACE_DIM != 2)
+    {
+        EXCEPTION("This function is only valid in 2D"); // LCOV_EXCL_LINE
```
mirams, line 1429: All `EXCEPTION`s should be covered as a rule, with a test that they throw the right exception, ideally?

fcooper8472: @mirams there were quite a few places that used to be:
```
assert(SPACE_DIM == 2); // LCOV_EXCL_LINE
```
that were changed to exceptions to satisfy the compiler about c_vector access.  So we're not ignoring anything more than we used to, it's just it was decided to move from assert -> exception.

PR #19, opened 2022-09-15
Adding suppressions to avoid 3 valgrind issues caused by switch from …

jmpf:
…MPICH to OpenMPI

These 3 suppressions should get Buildbot-style nightly memory testing back on track

### DIFF
--- a/chaste.supp
+++ b/chaste.supp
@@ -427,6 +427,40 @@
 }
 
 {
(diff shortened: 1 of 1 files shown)


fcooper8472 (approved):

fcooper8472:
Looks good! Go ahead and merge @jmpf

PR #23, opened 2022-10-21
Correctly use argument isFlatBottom in HoneycombVertexMeshGenerator constructor

### LINKED ISSUE #135: Correctly use argument isFlatBottom in HoneycombVertexMeshGenerator constructor
This issue corresponds to the old Chaste trac ticket [#3107](https://chaste.cs.ox.ac.uk/trac/ticket/3107).

To reiterate, the `HoneycombVertexMeshGenerator` constructor  accepts the argument `isFlatBottom` but ignores its value. This is in contrast to `CylindricalHoneycombVertexMeshGenerator`, where if `isFlatBottom` is true then the bottom boundary of the mesh is 'flattened'.

The task is therefore to correct the usage of `isFlatBottom` in the `HoneycombVertexMeshGenerator` constructor

georgiabeeton:
This pull request closes #135 by amending the `HoneycombVertexMeshGenerator` constructor to correctly use the argument `isFlatBottom`, with associated test `TestFlatBottomMesh()` in the test suite `TestHoneycombVertexMeshGenerator`.


mesh/src/vertex/HoneycombVertexMeshGenerator.cpp
```diff
@@ -34,6 +34,7 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
 #include "HoneycombVertexMeshGenerator.hpp"
+#include "CheckpointArchiveTypes.hpp"
```
mirams: I think this include should only feature in things actually making archives, so generally tests, and the odd "archiver" class?

mirams (changes requested):
Probably remove the `#include "CheckpointArchiveTypes.hpp"` from the class, and probably the test as well if it doesn't make archives.

mesh/test/vertex/TestHoneycombVertexMeshGenerator.hpp
```diff
@@ -39,6 +39,7 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <cxxtest/TestSuite.h>
 #include "HoneycombVertexMeshGenerator.hpp"
 #include "PetscSetupAndFinalize.hpp"
+#include "CheckpointArchiveTypes.hpp"
```
mirams: also this one, as I don't think archiving is being done in added code.

AlexFletcher: Resolved by commit 69c7ae2.

mesh/src/vertex/HoneycombVertexMeshGenerator.cpp
AlexFletcher: Resolved by commit 69c7ae2.

mirams (approved):

PR #30, opened 2022-12-08
#3102: Create a bug report template

kwabenantim:
## Summary
This pull request adds a bug report template to GitHub Issues.


MauriceHendrix (approved):
looks good to me!

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**System**
+Please provide relevant details of your system, such as:
+ * OS version
+ * CMake version
+ * C++ compiler version
+ * Relevant Chaste dependency versions e.g. PETSc, Boost, Sundials etc.
```
mirams, line 38: You could ask for the output of TestChasteBuildInfo.hpp here.

mirams (changes requested):
You could ask for the output of `TestChasteBuildInfo.hpp` which should give most of the info on dependencies and build type etc.

kwabenantim:
> You could ask for the output of `TestChasteBuildInfo.hpp` which should give most of the info on dependencies and build type etc.

Thanks @mirams. I've added some information about getting the output from `TestChasteBuildInfo.hpp` .

mirams:
Yeah, you don't want output on failure though, you want all the output when it passes as that will show the dependency versions

kwabenantim:
> Yeah, you don't want output on failure though, you want all the output when it passes as that will show the dependency versions

Thanks. It's now `ctest -V -R TestChasteBuildInfo` without the ~~`--output-on-failure`~~.

mirams (approved):

PR #45, opened 2023-01-11
Update AbstractCvodeSystem.cpp

MauriceHendrix:
I'm not sure what the best way of doing this is but I've just tried something (after Gary suggested it) that look slike had been tried before, so maybe just putting a comment about that will help.


mirams (changes requested):
Could you put your findings in an issue (that table, and code you ran) and link to it from the comment so that anyone investigating in future can see what we tried?

mirams (approved):

jmpf:
Issue #46

PR #56, opened 2023-01-17
Auatay ticket 2987 ode edges

BJackal:
Recent developments to the develop branch have been merged into the auatay branch for cell edge and interior data structures and use. Currently passes all Continuous tests locally and needs to be looked at for being merged into the development branch.


mesh/src/common/AbstractElement.cpp
```diff
@@ -51,6 +51,7 @@ AbstractElement<ELEMENT_DIM, SPACE_DIM>::AbstractElement(unsigned index, const s
       mIsDeleted(false),
       mOwnership(true),
       mpElementAttributes(nullptr)
+
```
fcooper8472: Remove spaces in this file

BJackal:
Required fixes:

1. Merge current develop for memory test fixes and to check for possible memory leaks. (Completed)

2. Move Edge based code within the AbstractMesh further down the Hierarchy into either VertexMesh or MutableVertexMesh. (Completed)

3. Create unique pointer for the EdgeHelper mEdges pointer such that its memory ownership is obvious and so that it will be deleted when moving out of scope. Allowing for deletion of unnecessary destructors. (In progress)

BJackal:
Pushed changes to the branch updating EdgeHelper to have a unique pointer to edges. So that it is explicitly known where this memory is held. Simillarly updated VertexMesh and MutableVertexMesh to now reference this information instead of copying it. Now fully builds again but has multiple failing tests related to VertexMesh.

fcooper8472:
This should hopefully resolve a number of the memory leaks. I'll check back on Monday after the memtest run finishes to see what's left to iron out.

fcooper8472:
Update after yesterday's fix: we're down to 5 tests with a memory leak. Making good progress - hoping to have the last memory leaks tracked down today.

@BJackal and @jmpf can we think about another code review meeting ASAP?

fcooper8472 (approved):
There is one outstanding issue (#106) but this is a good point to get it in to develop.

PR #62, opened 2023-01-30
Fix error in getting linux distribution info

bdevans:
This address the error detailed in this issue: https://github.com/Chaste/chaste-docker/issues/12 which is caused by the removal of `platform.linux_distribution()` in Python 3.7.

In doing so, it introduces a dependency on the `distro` package which still needs to be added to the list of Chaste dependencies if it is not already included.

It also updates the list of known macOS versions, removes the Python 2 boilerplate import and modernises the code to use f-strings.

### DIFF
--- a/cmake/Modules/ChasteHostOperatingSystem.py
+++ b/cmake/Modules/ChasteHostOperatingSystem.py
@@ -102,6 +102,14 @@ def get_mac_name_from_darwin_version(version_str):
         return 'macOS High Sierra'
     elif version_str.startswith('10.14.'):
         return 'macOS Mojave'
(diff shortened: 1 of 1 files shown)


fcooper8472 (approved):
Looks great, thanks @bdevans !

fieldofnodes:
Just to confirm:

docker run -it --init --rm -v chaste_data:/home/chaste chaste/develop

will get me the docker suitable for M1 arm64?

bdevans:
> Just to confirm:
> 
> docker run -it --init --rm -v chaste_data:/home/chaste chaste/develop
> 
> will get me the docker suitable for M1 arm64?

Correct!

Issue #68, opened 2023-02-23
TestGenerateSteadyStateCrypt fails on jammy


MauriceHendrix:
**Describe the bug**
TestGenerateSteadyStateCrypt fails on ubunto jammy but passes on bionic.
'''
test 326
    Start 326: TestGenerateSteadyStateCrypt

326: Test command: /home/uczmh2/develop/chaste_build/crypt/test/TestGenerateSteadyStateCrypt
326: Test timeout computed to be: 10000000
326: Running 1 test
326: 
326:  ***** TestGenerateSteadyStateCrypt.hpp *****
326: 
326: 
326: Entering TestGenerateSteadyStateCryptArchives
326: 
326: /home/uczmh2/develop/Chaste/crypt/test/simulation/TestGenerateSteadyStateCrypt.hpp:175: Error: Expected (expected_cell_count_lb < p_simulator->rGetCellPopulation().GetNumRealCells()), found (425 >= 422)
326: Failed
326: 
326: Failed 1 of 1 test
326: Success rate: 0%
1/1 Test #326: TestGenerateSteadyStateCrypt .....***Failed  1248.60 sec
```

PR #69, opened 2023-02-23
Update TestGenerateSteadyStateCrypt.hpp

MauriceHendrix:
Further relax number of cells check, due to variations in compiler / use of CVODE and differences in dependancies.
see #68


kwabenantim (approved):
I have confirmed that dropping the lower bound from 425 to 420 causes `TestGenerateSteadyStateCrypt` to pass on jammy. It continues to pass on focal as well.

Issue #59, opened 2023-01-17
Create private Chaste repo with associated wiki


AlexFletcher:

AlexFletcher:
See [here](https://docs.google.com/document/d/1_Rhs_f8GHJLQLJ63wfSG9C1NneAz-EEqA4ktAQ9Lw68/edit?usp=sharing) for a list of old wiki pages. We will annotate those pages that should be moved to the public Chaste website and those which should be moved to the private developer wiki. If you don't have access to the google doc and think you should have, then contact me.

AlexFletcher:
In the google doc linked to in my above comment, I've now colour coded those wiki pages that I *think* should be copied over to either the new public Chaste website or to the new private developer wiki. Please could someone (@fcooper8472 @mirams @jmpf ?) take a quick look through and either let us know if they agree, or amend as necessary.

mirams:
Looks good to me, tweaked a couple

AlexFletcher:
Thanks @mirams. Any comments @fcooper8472 @jmpf or anyone else?

jmpf:
Read it and it looks reasonable.

I'd quite like an archive of stuff like BuildServerSetup, but I think that archiving the entire wiki somewhere safe before it dies is the way to do it.

fcooper8472:
@jmpf the entire archive already exists.  See, e.g. [here](https://github.com/Chaste/trac_archive/wiki/Build-Server-Setup) for the page you mentioned.

mirams:
Be nice to get all the wiki links working, the top one there does but the rest (nested under InstallGuides/* ??) don't seem to be working?

fcooper8472:
Yes, that should be fairly straightforward to fix. This is just down to the fact that GitHub wiki page titles are determined by their file names.

Fix (hopefully...) incoming.

fcooper8472:
@mirams I've got a fix ready for a lot of the links, but there are quite a few that don't seem to make any sense.

I'll wait to push these changes until I can get back into the old wiki to investigate what the links originally looked like.

fcooper8472:
Ah - think I see what's happened.  Links to class names in tutorials etc have been erroneously changed to wiki links. That's a bit of a pain!

fcooper8472:
@mirams and @jmpf this should now all be resolved.

The class names now direct to the appropriate public docs page (e.g. https://chaste.cs.ox.ac.uk/public-docs/classAbstractForce.html) and the genuine wiki links should all now work.

[...]

PR #73, opened 2023-03-01
Removing scons and adding an error message

MauriceHendrix:
This removes Scons code, except for the error message thrown when you try to build with Scons.
However I stilll need to know where to point to, which relates to #59


jmpf (approved):
Deleted files are now redundant so PR approved.

- SConstruct error message is a placeholder and should (later) be a real URL to the Wiki
- Subfolders python/hostconfig and python/hostconfig/machines can now be removed too

jmpf:
Also todo: remove mentions of Scons in code comment instructions.

fcooper8472:
This looks good to me.

fcooper8472 (approved):

chaste-lucid.supp
kwabenantim, line 1: I think the remaining text looks like it may be missing something. Maybe we can consider removing the rest of the comment. I don't know if the flag is necessary for CMake or not @MauriceHendrix .

```
#################################################
# Note if memory leaks depend on the context
#################################################
```

chaste.supp
kwabenantim, line 1: Maybe worth removing the rest of the comment here too.

```
#################################################
# Note if memory leaks depend on the context
#################################################
```

chaste-lucid.supp
MauriceHendrix: I think that was scons related indeed

global/test/TestCommandLineArguments.hpp
```diff
@@ -54,13 +54,6 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  *
  *
- *  For example:
```
mirams, line 57: This really wants replacing with cmake example rather than just deleting

apps/texttest/weekly/MeshConvert/Tutorial.wiki
```diff
@@ -15,19 +15,6 @@ our special binary version of the Triangle/Tetgen format.
  to containing element indices.  The presence of this file cuts the calculation cost at mesh load time.
  * If a Chaste format fibre file (.axi or .ortho) is associated with the mesh then this will also be converted to our binary format.
 
-=== Making the executable(s) ===
```
mirams, line 18: This really wants replacing with cmake example rather than just deleting

MauriceHendrix: I thought naking standalone excutables wasn't possible anymore? @fcooper8472 ? unless I have missunderstood?

mirams: We don't try and make them for people due to difficulties getting compatible cross-platform ones for Linux (used to generate the cardiac executable for download) but still should be able to make them (like ApPredict).

fcooper8472: Yes, it's still absolutely possible. We no longer ship a cardiac executable that you can expect to run portably without compiling it yourself, but the top level apps directory (and indeed the apps directory in any user project) can have standalone executables with their own `int main()` methods.

MauriceHendrix: ok but how do we write that in the tutorial?

fcooper8472: This would just be the corresponding instructions but using CMake, I suppose?
```
scons chaste_libs=1 compile_only=1 exe=1 build=GccOpt apps
```
would become, assuming you have configured Chaste
```
make Chaste
```
or
```
make MeshConvert
```
and the `build=GccOpt_ndebug` equivalent would be passing `-DCMAKE_BUILD_TYPE=Release` at configure time.

MauriceHendrix:
addressing the new comments about build properties for cmake in #94

PR #75, opened 2023-03-08
Update TestCodegen.hpp

MauriceHendrix:
Extra codegen checks are only enabled in debug mode so only tested in debug


heart/test/ionicmodels/TestCodegen.hpp
```diff
@@ -180,17 +182,21 @@ class TestCodegen : public CxxTest::TestSuite
 
         // Optimised model
         // test error from codegen about concentration
+#ifndef NDEBUG
         Cellnegative_concentration_paci_hyttinen_aaltosetala_severi_ventricularVersionFromCellMLOpt opt_model(p_solver, p_stimulus);
         TS_ASSERT_THROWS_CONTAINS(opt_model.Compute(0.0, end_time, sampling_interval), "Concentration JSR_calcium_concentration below 0");
+
```
mirams: # endif here?

fcooper8472 (changes requested):
Looks fine to me as once you've added the missing endif!

PR #80, opened 2023-03-15
make chaste exception inhert from std::runtime_error

MILeach:
Makes Exception inherit from std::runtime_error so that uncaught exceptions cause the error message to be printed. At present, the user is just informed that an exception was thrown, but is not shown the error message. Uses of catching exceptions by value have been replaced with catching by reference, this is necessary to handle the now-polymorphic Exception type properly and is also good practice.


global/src/Exception.hpp
```diff
@@ -72,6 +73,11 @@ class Exception
      */
     Exception(const std::string& rMessage, const std::string& rFilename, unsigned lineNumber);
 
+    /**
+     * Virtual destructor
+     */
+    virtual ~Exception();
```
fcooper8472: can this be `= default`?

fcooper8472:
@MILeach I've just rebased this onto develop to see how it's looking with the new CI checks.

Any red crosses should be legitimately red now!

fcooper8472:
@MILeach just looked at this PR and it looks fine to me. I've just done a spot of tidying in the class unrelated to your changes. But I'm approving the PR so please feel free to merge assuming it all stays green!

fcooper8472 (approved):

mirams (approved):
Looks good, can we write something up somewhere about what the benefit of this is? Release notes maybe?

jmpf:
Changes to header file naming in #100 (boost::filesystem to std::filesystem) caused conflict in merge.  Should be easily resolved...

jmpf:
Just for the record, I made this change to the code:
```
diff --git a/global/test/TestWarnings.hpp b/global/test/TestWarnings.hpp
index 6a0c6f11a..4335312fb 100644
--- a/global/test/TestWarnings.hpp
+++ b/global/test/TestWarnings.hpp
@@ -140,6 +140,7 @@ public:
     void TestLastTestWithWarningsIsNoisy() //Needs to happen last (after any QuietDestroy()), so that a warning is printed
     {
         TS_ASSERT_THROWS_THIS(Warnings::Instance()->GetNextWarningMessage(),"There are no warnings");
+        Warnings::Instance()->GetNextWarningMessage();

```

On my machine with this pull request:
```
Entering TestLastTestWithWarningsIsNoisy

/home/jmpf/Workspace/Chaste/global/test/TestWarnings.hpp:140: Error: Test failed: 
Chaste error: ./global/src/Warnings.cpp:127: There are no warnings
```
but I also get the same (i.e. an informative error) in develop ***before*** the pull request.  Go figure.

jmpf (approved):
Now safe to merge

Issue #84, opened 2023-03-16
Fix cardiac electromechanics -np3 uncaught exception on  22.04 LTS


jmpf:
**Describe the bug**
Test TestAbstractContractionCellFactory fails in parallel -np3 on newchaste server which is running 22.04 LTS (Jammy).  Also pertinent is that it compiles against PETSc 3.15 with g++ 11.3.

**To Reproduce**
See https://github.com/Chaste/Chaste/actions/runs/3938422099/jobs/6737088563

**Expected behaviour**
This test has an uncaught exception.
terminate called after throwing an instance of 'Exception'

**System**
 * OS Ubuntu 22.04 .1 LTS
 * C++ compiler version g++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
 * PETSc 3.15.5

jmpf:
On newchaste (22.04) TestAbstractContractionCellFactory is failing with -np3 when it hits a KSP_DIVERGED_BREAKDOWN in a PETSc KSP solve.

On Scoop I ran with `mpirun -np 3 ./heart/test/TestAbstractContractionCellFactory -ksp_monitor` to investigate what is supposed to happen on a previous configuration (Ubuntu 20.04.4, g++ 9.4.0, PETSc 3.12.5).
![OldConfig](https://user-images.githubusercontent.com/3852505/225607360-bb6fd7a5-4327-46ba-b5ec-5f713365ab1f.png)
Each point is shows which iteration a KSP solver is on.  There are multiple calls to KSP solvers that converge within 10 iterations but, within the main nonlinear solver there are about 10 calls to a solver that takes many steps to converge.  The -np3 version converges more slowly:
Sequential:
`141 KSP Residual norm 1.419166734049e-10 
142 KSP Residual norm 6.724390967265e-11 
Passed
`
-np3:
`452 KSP Residual norm 1.342054903103e-10 
453 KSP Residual norm 1.181856560959e-10 
Passed
`

jmpf:
On newchaste (before the exception) the trace looks similar.  Convergence is slightly faster until the final step of the  big nonlinear solve.  The exception happens at iteration 299 of the KSP solver.
![NewConfig](https://user-images.githubusercontent.com/3852505/225609837-2c1d8461-60a4-4c22-875c-603d45fdd966.png)

jmpf:
[...]

PR #86, opened 2023-03-17
#84 Stop nonlinear solver falling over on PETSc 3.15

jmpf:
This stops the code from failing (and the tests in TestCardiacElectroMechanicsProblemLong all pass!).  However the new PETSc 3.15 which is converging poorly takes more than twice as long to complete the suite (11m15s versus 4m15s).

### DIFF
--- a/continuum_mechanics/src/solver/AbstractNonlinearElasticitySolver.hpp
+++ b/continuum_mechanics/src/solver/AbstractNonlinearElasticitySolver.hpp
@@ -1810,26 +1810,25 @@ double AbstractNonlinearElasticitySolver<DIM>::TakeNewtonStep()
     // warn if ksp reports failure
     KSPConvergedReason reason;
     KSPGetConvergedReason(solver,&reason);
-
-    if (reason != KSP_DIVERGED_ITS)
+    if (reason == KSP_DIVERGED_ITS || reason == KSP_DIVERGED_BREAKDOWN)
     {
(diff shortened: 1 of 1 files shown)


continuum_mechanics/src/solver/AbstractNonlinearElasticitySolver.hpp
```diff
@@ -1832,6 +1822,15 @@ double AbstractNonlinearElasticitySolver<DIM>::TakeNewtonStep()
         WARN_ONCE_ONLY("Linear solve (within a mechanics solve) didn't converge, but this may not stop nonlinear solve converging")
         // LCOV_EXCL_STOP
     }
+    else
+    {
+        // Throw an exception if the solver failed for any reason other than DIVERGED_ITS.
```
MauriceHendrix, line 1827: In that case I guess we shuld exclude it from coverage? @fcooper8472

MauriceHendrix (approved):

MauriceHendrix:
With this change TestCardiacElectroMechanicsProblemLong also passes on mysetup too now.

continuum_mechanics/src/solver/AbstractNonlinearElasticitySolver.hpp
jmpf: Coverage ought to be unchanged by this commit.

PR #89, opened 2023-03-22
Fix portability workflow

kwabenantim:
- This fixes the path to the checkout directory in the Portability GitHub Actions workflow:
https://github.com/Chaste/dependency-modules/actions/runs/4501485360


kwabenantim:
@fcooper8472 as portability is running on PRs now, do we need to remove the cron job?

fcooper8472 (approved):

PR #94, opened 2023-04-04
Added comments about build options

MauriceHendrix:
Added comments siuggested in closed PR: #73


global/test/TestCommandLineArguments.hpp
```diff
@@ -54,6 +54,9 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  *
  *
+ *  For example:
```
mirams: This `CommandLineArguments` is to do with passing in arguments to compiled executables. scons used to have an option to "build and run with these command line arguments to the test/executable" but I guess that would now be something to do with `ctest` if that can actually do it?

mirams (changes requested):

global/test/TestCommandLineArguments.hpp
MauriceHendrix: Probably, but worth checking with @fcooper8472

mirams:
For this to be sorted we want to give advice on how to pass command line arguments to a test executable at the time you write `ctest TestBlah....` 

Some stackoverflow here that looks very complicated! https://stackoverflow.com/questions/28812533/how-to-pass-command-line-arguments-in-ctest-at-runtime

Alternatively, we don't ever do this anymore as you can manually run the test executable and do it directly?

apps/texttest/weekly/MeshConvert/Tutorial.wiki
```diff
@@ -15,6 +15,15 @@ our special binary version of the Triangle/Tetgen format.
...
  * If a Chaste format fibre file (.axi or .ortho) is associated with the mesh then this will also be converted to our binary format.
 
+
+=== Making the executable(s) ===
+We no longer distribute cardiac executables due to difficulties getting compatible cross-platform executables. 
+However the top level apps directory (and indeed the apps directory in any user project) can have stand-alone executables with their own int main() methods.
+to compile these as an executable:
+
+`cmake .. where .. is the source folder for Chaste` followed by
+`make <target name> -DCMAKE_BUILD_TYPE=Release` where is your build tareget e.g. Chaste, or a user project such as ApPredict
```
mirams:
```suggestion
`make <target name> -DCMAKE_BUILD_TYPE=Release` where is your build target e.g. Chaste, or a user project such as ApPredict
```

mirams (approved):

Issue #95, opened 2023-04-24
Exception constructing NodesOnlyMesh with certain sized boxes


NicholasFan235:
**Description**
Issue when initialising DistributedBoxCollection in NodesOnlyMesh for Node based simulations.

Error when initialising NodesOnlyMesh using NodesOnlyMesh::ConstructNodesWithoutMesh()
If the corners of the box it tries to create are (2,2) and (200,200) it crashes.
Someone mentioned possibly the box size can't be a multiple of the interaction distance (I'm using 1.5).
But it might be a bit more complex because NodesOnlyMesh::SetUpBoxCollection() offsets the boundary by 1e-14 to ensure it's bigger.
Possible floating point precision error, but unlikely?

**To Reproduce**
Created minimal unit test which shows behaviour.
Test places 4 cells (one in each corner). Depending on their position can cause a "The point provided is outside all the boxes" exception from DistributedBoxCollection.
Similar error to Issue #22 

Test is meant to go in src/test/mesh folder

```c++
/*

Copyright (c) 2005-2023, University of Oxford.
All rights reserved.

University of Oxford means the Chancellor, Masters and Scholars of the
University of Oxford, having an administrative office at Wellington
Square, Oxford OX1 2JD, UK.

This file is part of Chaste.

Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are met:
 * Redistributions of source code must retain the above copyright notice,
   this list of conditions and the following disclaimer.
 * Redistributions in binary form must reproduce the above copyright notice,
   this list of conditions and the following disclaimer in the documentation
   and/or other materials provided with the distribution.
 * Neither the name of the University of Oxford nor the names of its
   contributors may be used to endorse or promote products derived from this
   software without specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
[...]

PR #99, opened 2023-05-16
#95 Make sure that initial bounding box of NodesOnlyMesh is always sw…

jmpf:
…ollen relative to its size


jmpf:
Pull request should fix #95

fcooper8472 (approved):
Looks as though some tests are spuriously failing due to not correctly downloading some dependencies.

Looks fine to me!

jmpf:
> Looks as though some tests are spuriously failing due to not correctly downloading some dependencies.

Yeah.  That's what I thought.  Perhaps some ubuntu repository was offline at the time.

PR #104, opened 2023-06-08
103 bbr wp 41 barcelona

### LINKED ISSUE #103: BBR WP 4.1 Add some new cell-based examples for use in Barcelona workshop
We would like to add a few new examples of cell-based functionality to the trunk. These include:

- a new stochastic cell division rule based on the [von Mises-Fisher distribution](https://en.wikipedia.org/wiki/Von_Mises%E2%80%93Fisher_distribution);
- a new cell cycle model, similar to BernoulliTrialCellCycleModel, but with a prescribed dependence of mDivisionProbability on spatial location;
- a new simulation modifier that pins cells along the left-hand boundary and imposes a fixed strain rate from the right-hand boundary, similar to [Finegan et al (2019)](https://doi.org/10.1371/journal.pbio.3000522).

AlexFletcher:
This pull request relates to issue [103 ](https://github.com/Chaste/Chaste/issues/103) and introduces the following new functionality into the cell-based code, for the Barcelona workshop:

- VonMisesVertexBasedDivisionRule, a new class that allows the user to specify the parameters of a von Mises (circular) distribution for the angle of cell division in vertex models - here, cells divide with random orientations but with some preferred mean and some measure of variation about this mean;
- BiasedBernoulliTrialCellCycleModel, a new class that is similar to the existing class BernoulliTrialCellCycleModel but which allows the division probability per unit time to vary linearly across a specified axis of the tissue - this is implemented by storing the bias in division probability in CellData;
- DivisionBiasTrackingModifer, a new class that is similar to the existing class VolumeTrackingModifier but which instead store's each cell's bias in division probability based on its location parallelt to a specified axis of the tissue - this is a helper class for BiasedBernoulliTrialCellCycleModel, similar to how VolumeTrackingModifier is a helper class for ContactInhibitionCellCycleModel;
- ConstantTargetAreaModifier, a new class that is similar to the existing class SimpleTargetAreaModifier but allows *no* growth in cell areas over time (force classes such as FarhadifarForce require the user to provide a target area modifier, hence the need for this in the case where the user doesn't actually want to model cell growth);
- SlidingBoundaryCondition, a new class that forces boundary nodes on the left-hand boundary of a tissue to stay at their initial x-coordinate, but allows them to 'slide' up and down in the y direction;
- additional tests and test suites associated with these new classes.

### DIFF
--- a/cell_based/src/cell/cycle/BiasedBernoulliTrialCellCycleModel.cpp
+++ b/cell_based/src/cell/cycle/BiasedBernoulliTrialCellCycleModel.cpp
@@ -1,0 +1,132 @@
+/*
+
+Copyright (c) 2005-2023, University of Oxford.
+All rights reserved.
+
+University of Oxford means the Chancellor, Masters and Scholars of the
+University of Oxford, having an administrative office at Wellington
+Square, Oxford OX1 2JD, UK.
+
+This file is part of Chaste.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice,
+   this list of conditions and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+ * Neither the name of the University of Oxford nor the names of its
+   contributors may be used to endorse or promote products derived from this
+   software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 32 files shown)


AlexFletcher:
I've just pushed the following additional new functionality into the cell-based code, for the Barcelona workshop:

- LabelDependentBernoulliTrialCellCycleModel, a new class that is similar to the existing class BernoulliTrialCellCycleModel but which allows the division probability per unit time to varyaccording to whether a cell is labelled;
- PlanarPolarisedFarhadifarForce, a new class that is similar to the existing class FarhadifarForce but which has a line tension term that varies depending on the orientation of the edge relative to the x-axis;
- ExtrinsicPullModifier, a new class that imposes a constant strain rate either on the right-most nodes in a cell population or on all of them, towards the right;
- additional tests and test suites associated with these new classes.

This completes the new functionality that we plan to need for the Barcelona workshop (except for @BJackal 's major pull request that @fcooper8472 and @jmpf are already dealing with).

fcooper8472 (approved):
Looks good to me!

PR #111, opened 2023-06-15
Add subcellular element model

### LINKED ISSUE #110: Add subcellular element model
As part of the BBR grant, we should add the subcellular element model to the core chaste library.

# Existing Code

This work builds on partially complete code initially developed in 2013. 

The [**original copy**](https://chaste.cs.ox.ac.uk/trac/browser/chaste/projects/Sem) of the code is available on trac 
A [**working copy**](https://github.com/Chaste/sem-project) is available in the [Chaste/sem-project](https://github.com/Chaste/sem-project) repository

# Development Tasks

- [ ] Update the existing project code in the [**working copy**](https://github.com/Chaste/sem-project) to build against current develop
- [ ] Add new required functionality
- [ ] Add tests for new functionality
- [ ] Ensure subcellular element code meets required standards & is sufficiently tested
- [ ] Create tutorial & documentation

MILeach:
See below for status drafted by @fcooper8472


fcooper8472:
# Subcellular Element Model

Implements the Subcellular Element Method (SEM) from Sandersius & Newman (2008) *Phys. Biol.* 5 015002 as a first-class Chaste cell population. This PR closes #110.

## What this adds

Each biological cell is represented by a `SemElement` - a collection of subcellular nodes interacting through a modified Morse potential. Multiple cells interact via the same potential with separately tunable parameters, producing emergent cell shapes, cortical mechanics, and collective motion.

**Mesh layer** (`mesh/src/sem/`)
- `SemElement`, `SemMesh` - templated 1/2/3D mesh; neighbour queries via `DistributedBoxCollection`
- `SemMeshReader` / `SemMeshWriter` - `.node`/`.sem` file I/O and VTK output; surface extraction via Delaunay triangulation (`SemElementGeometry`) for 3D visualisation in Paraview
- `SemSingleElementMeshGenerator` / `SemMultiElementMeshGenerator` — regular-grid mesh construction in 1/2/3D with automatic interior/boundary node classification

**Cell population** (`cell_based/src/population/`)
- `SemBasedCellPopulation` - maps one `CellPtr` to one `SemElement`; full lifecycle hooks implemented

**Force laws** (`cell_based/src/population/forces/`)
- `SemForce` - full modified Morse potential (Sandersius & Newman eq. 3)
- `SemLinearForce` - harmonic spring approximation (eq. 4); equivalent near equilibrium, cheaper to evaluate
- `SemRegionalForce` -  node-region-dependent spring with separate stiffness and rest-length for interior and boundary/cortex nodes; fully tested and serialised
- `SemGaussianRandomForce` - independent per-node Gaussian noise, correctly scaled to the overdamped Langevin correlation ⟨ξ^m ξ^n⟩ = 2Dη²δ(t−t')
- `SemSpatiallyCorrelatedRandomForce` - spatially correlated noise via `OffLatticeRandomFieldGenerator`
- `SemParameterScaling.hpp` - `SemComputeNScaledParameters<DIM>()` implements the N-dependent scaling laws from Section 2 of the paper; `ApplyNScaledIntraParameters()` / `ApplyNScaledInterParameters()` convenience methods on both force classes ensure physically consistent parameters as node count N is varied

## Status

This is a stable, tested base suitable for running SEM simulations. All 7 tests pass:

`TestSemMesh`, `TestSemMeshWriter`, `TestSemBasedCellPopulation`, `TestSemParameterScaling`, `TestForces`, `TestSemBasedSimulation`, `TestRunningSemBasedSimulationsTutorial`

The tutorial (`TestRunningSemBasedSimulationsTutorial`) walks through three complete simulations - a 2D single cell, a 2D multi-cell system, and a 3D single cell - covering both uncorrelated and spatially correlated noise, with Paraview visualisation instructions.

## Known limitations

A small number of correctness issues remain before this PR is marked ready for review. The most significant is a read/write round-trip bug in `SemMeshReader` (header field mismatch) and the absence of an archive/restart test. Neither affects forward simulation.

## Related issues

- Closes #110
- Closes #443 - `SemRegionalForce` is fully implemented, tested, and serialised with `OutputForceParameters` in place
- Closes #445 - `SemForce` (Morse) and `SemLinearForce` (harmonic) both handle intra- and inter-cellular interactions with separate parameter sets; the original stub classes `SemIntraCellularForce` / `SemInterCellularForce` have been removed

fcooper8472:
### Update: physics fidelity, persistence, robustness, force API, and VTK output

Since the last `develop` merge, this adds a round of fixes and API work on the SEM implementation, checked against Sandersius & Newman (2008). All new/changed code is tested; the full SEM `mesh` + `cell_based` continuous suites pass and build warning-clean under `-Wall -Wextra`.

**Physics fidelity**
- N-scaled damping is now returned from the force scaling helpers, so `η = η₀N` can no longer be silently dropped — `ApplyNScaled{Intra,Inter}Parameters` return the full scaled parameter set. (1e5b9064d)
- `SemBasedCellPopulation::CheckForStepSizeException` implemented: throws a `StepSizeException` with a suggested smaller timestep when a node would move further than the mesh's maximum interaction distance, instead of silently exploding. (803d7c13e)

**Mesh persistence — round-trip & restart**
- `SemMeshReader` validates the mesh-file dimension against its `DIM`. (17d6706fa)
- Fixed the write→read round-trip: read *all* element nodes (previously truncated to `DIM+1`) and persist per-node regions. (7fb896360)
- Archive/restart: rebuild the box collection on load and archive the SEM scalar members. (b7212b5ec)
- Removed the vestigial `SemElement` cell-id field (unused; not required for division). (f76722ad3)

**Robustness**
- Guard against empty node→element membership in `SemForce`/`SemLinearForce`. (ca35090b4)
- Fixed a 1D `SemMultiElementMeshGenerator` out-of-bounds element-spacing access. (7f52acd64)
- Removed the legacy `SemMeshGenerator` (unused — its constructor built an empty mesh), which also disposes of two latent bugs and its redundant PyChaste wrapper. (046da0b40)

**Force API**
- Reworked `SemRegionalForce`: identical harmonic mechanics to `SemLinearForce`, but with per-region spring constant and rest length for an arbitrary number of regions (mean-combined per node pair), a configurable cut-off, full setters, and harmonised to a single `<DIM>` template. (324f699a2)

**VTK output**
- Per-element scalars can now be written as VTK **cell** data: `SemMeshWriter::AddElementData` expands a per-element payload onto every VTK cell of that element (point cloud + reconstructed surface), and `SemBasedCellPopulation::WriteVtkResultsToFile` emits cell-writer outputs (e.g. cell volume) and `CellData` items this way. The tutorial is updated to demonstrate colouring cells by these arrays in Paraview. (10ab52694, 9325d25fe)

**Still outstanding** (not in this batch): periodic boundary conditions (shelved for now), a close-packed / equilibrated initial-condition generator, and cell division.

fcooper8472:
### Update: lattices, spherical cells, applied loads, PDEs, and PyChaste

Since the last update this adds close-packed lattices and a spherical cell generator, a uniaxial load force with a rheology tutorial, the PyChaste wrappers, and a coverage pass. 24 commits, +6149/-343 across 43 files. The full SEM `mesh` + `cell_based` continuous suites pass.

**Lattices and cell geometry**
- `SemLatticeGeometry` generates lattice positions for both the nodes within an element and the elements within a multi-element mesh, offering cubic and close-packed arrangements via the new `SemLatticeType`. Close packing gives a triangular lattice in 2D and HCP in 3D, with nearest neighbours exactly one spacing apart. (00dae64bb, bfc386a2e)
- `SemSphericalElementMeshGenerator` carves a ball of N nodes out of a close-packed lattice, so a cell starts spherical rather than as a cube. Node spacing follows the paper's `r_eq(N) = 2R(p/N)^(1/3)`, verified against the FCC packing density in a simulation test. Boundary nodes are labelled by depth from the surface. (eb8f5a76f, 52ab637e8)
- The generators no longer set a node radius that nothing read. (c3777c9f5)

**Applied loads and rheology**
- `UniaxialLoadForce` compresses or stretches a specimen along an axis by loading two slabs of nodes at opposite ends. Slabs are identified once and then held fixed, so the same nodes stay loaded as the specimen deforms, and the applied forces sum to zero. (a8d40093e)
- SEM random forces can cool towards zero over a time window, which lets a specimen be melted and then settled before a measurement begins. (d0ecf8f3f)
- `CellPopulationExtentWriter` records the per-axis radius of gyration, used to measure strain. A radius of gyration is used rather than a bounding box because a bounding box is fixed by whichever two nodes happen to lie furthest apart, which makes it both noisy and sensitive to lattice orientation — a ball carved from a lattice can report a markedly anisotropic bounding box while being genuinely isotropic. (251fb540f)
- A single-cell rheology tutorial putting these together. (be0cde34e)

**Coverage and dead code**
- Remaining test stubs replaced with real coverage, and the gaps closed across the SEM forces, mesh geometry and population classes. Everything now reachable is covered; the handful of lines that are not carry `LCOV_EXCL_LINE` with a stated reason. (429221edf, 6757fd00a, 992b8b776, 360524a07, ce5f2d291)
- Working through those gaps turned up code that was unreachable rather than untested, now removed: `SemMesh::GetBoxCollection` (no callers), the face-reading loop in `ConstructFromMeshReader` (`SemMeshReader::GetNumFaces()` is hardcoded to zero), the node-attribute branch (`SemMeshReader` never populates them), `SemMeshWriter`'s node map (only reachable when `IsMeshChanging()`, which a `SemMesh` never sets), and two `Solve*Mapping` methods that overrode nothing. (027f85205, c58399684)
- One left in place for review: every throwing check in `SemBasedCellPopulation::Validate()` is also unreachable, since the constructor requires exactly one cell per element and rejects duplicate or out-of-range indices, and the population iterator skips cells on deleted elements. Worth simplifying, but not in this PR.

**PDE modifiers**
- Box domain PDE modifiers work with a SEM population as-is, and are now tested. They generate their own finite element mesh over a supplied cuboid, so all they need from the population is each cell's centre, which SEM supplies as its element centroid. (50d26ea0f)
- Growing domain modifiers now throw instead of returning a null mesh that the caller dereferences without checking. They need a finite element mesh built from the population, which needs a triangulation of each element's node cloud — and unlike a `VertexElement`, whose nodes form an ordered ring that can be fanned into triangles about its centroid, a `SemElement` holds an unordered cloud, and SEM cells are not space-filling, so there is no conforming tissue mesh to build. (9a9c02e2e)

**PyChaste**
- Wrappers for the new SEM classes, `UniaxialLoadForce` and `CellPopulationExtentWriter`. (bfa6a1e2b, a838abdf6)
- `SemLatticeType` is now a struct wrapping an enum rather than a bare enum, which is the only form the wrapper generator can bind — `RelativeTo` in `FileFinder.hpp` is wrapped the same way. This was failing all three PyChaste CI jobs. Opened Chaste/cppwg#113 to add proper enum support upstream, after which this can go back to a plain enum. (a4d0ac7c7)

Issue #100, opened 2023-05-22
Boost Filesystem fails to copy between volumes.  Consider replacing with std::filesystem.


jmpf:
**Describe the bug**
Boost Filesystem fails to copy between volumes.  This is a known Boost bug: https://stackoverflow.com/questions/24209886/invalid-cross-device-link-error-with-boost-filesystem

**To Reproduce**
This may be specific to Boost 1.74.0 with /tmp and /home on different volumes:
```
jmpf@csu15256:~/Workspace/Chaste/build$ df /tmp/
Filesystem     1K-blocks      Used Available Use% Mounted on
/dev/nvme0n1p2 490617784 404090124  61532180  87% /
jmpf@csu15256:~/Workspace/Chaste/build$ df /home
Filesystem     1K-blocks      Used Available Use% Mounted on
/dev/nvme1n1p1 491133848 330703932 135408220  71% /home
```
```
ctest -VV -R TestFileFinder
...
13: Entering TestFindMatches
13: Passed
13: Entering TestCopying
13: 
13: /home/jmpf/Workspace/Chaste/global/test/TestFileFinder.hpp:362: Error: Test failed: 
13: Chaste error: ./global/src/FileFinder.cpp:332: boost::filesystem::copy_file: Invalid cross-device link: "/home/jmpf/Workspace/Chaste/global/test/TestFileFinder.hpp", "/tmp/jmpf/testoutput/TestFileFinder_TestCopying/TestFileFinder.hpp"
13: Failed
13: Entering TestDefaultConstructor
13: Passed
```

**Workaround**
Change environment variable to give a different output location
```
export CHASTE_TEST_OUTPUT="/home/jmpf/tmp-chaste/"
```

**Expected behaviour**
Test should pass

**System**
Please provide relevant details of your system, such as:
 * OS version: Ubuntu 22.04
 * Relevant Chaste dependency versions Boost: libboost-filesystem-dev:amd64              1.74.0.3ubuntu7

jmpf:
We currently use -std=c++14 for compiling Chaste.

It's not clear if all the versions of C++ compilers in use will support the std::filesystem.  Probably they all will but some may need `#include <experimental/filesystem>` rather than  `#include <filesystem>` .

```
jmpf@csu15256:~$ cat temp.cc
#include <filesystem>
namespace fs=std::filesystem;

int main(){}
jmpf@csu15256:~$ g++ -std=c++17 temp.cc 
[...]

PR #115, opened 2023-06-23
Issue 100

jmpf:
Fixes issue #100 on machine with /tmp and /home on different SSDs.
Should be ready to go.


fcooper8472:
There's a compile failure with the classic intel compiler. But the warnings it's emitting have prompted me to open #116 (and PR #117).

I think removing intel classic is the right move at this juncture, and will have the added benefit of fixing this specific build error.

fcooper8472 (approved):
Content looks good - happy for this to be merged as soon as the CI is green.

jmpf:
Slightly different behaviour has altered coverage.  Investigating.

jmpf:
Note that ddb0ecb removes corner case code for creating a directory with a trailing slash.  Without this the old code would fall over like this:
`***** TestArchivingHelperClasses.hpp *****


Entering TestArchiveLocationInfoMethods
Passed
Entering TestArchiveLocationInfoProcessUniqueNaming
Passed
Entering TestProcessSpecificArchive
Passed
Entering TestArchiveOpenerReadAndWrite

Chaste termination: /home/jmpf/Workspace/Chaste/global/src/OutputFileHandler.cpp:231: Error making test output folder: boost::filesystem::create_directory: Permission denied: "/tmp/jmpf/testoutput/archiving_helpers/."`

The std::filesystem doesn't need this

PR #117, opened 2023-06-23
Remove classic intel compiler from all workflows

### LINKED ISSUE #116: Retire classic Intel C++ compiler
We currently compile all chaste libs with:

1. clang-12
2. clang-14
3. gcc-10
4. gcc-12
5. intel-classic
6. intel-llvm

The latest version of intel-classic has started emitting the following warning:

```
icpc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '-diag-disable=10441' to disable this message.
```

Presumably at some point in the next few months the binaries will simply not be available from the intel deb repository that we currently get them from.

I propose getting rid of the classic intel compiler at this point, since it's been superseded by the new version (which we are thoroughly testing).

fcooper8472:
See #116

### DIFF
--- a/.github/workflows/ubuntu-compilers.yml
+++ b/.github/workflows/ubuntu-compilers.yml
@@ -33,9 +33,6 @@ jobs:
           - cc: "gcc-12"
             cxx: "g++-12"
             build-type: "Debug"
-          - cc: "icc"
-            cxx: "icpc"
-            build-type: "Debug"
           - cc: "icx"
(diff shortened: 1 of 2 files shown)


jmpf (approved):
Yep.  Turn it off.

Issue #122, opened 2023-06-26
Make chaste_codegen version appear in autogenerated Version.cpp ?


mirams:
Whereas other executables write their versions into the build-time-generated `Version.cpp`, chaste_codegen doesn't and does some weird looking file opening to read it. Think about whether we can put in the right info at build time.
```
const char* ChasteBuildInfo::GetCompilerVersion()
{
    return "9.4.0";
}

std::string ChasteBuildInfo::GetChasteCodegenVersion(){
    FileFinder chaste_build_root("", RelativeTo::ChasteBuildRoot);
    char buffer[100];
    std::string codgen_version_command = chaste_build_root.GetAbsolutePath() + "/codegen_python3_venv/bin/chaste_codegen --version";
    FILE* codegen_version = popen(codgen_version_command.c_str(), "r");
    std::string results = "";

    while (fgets(buffer, 100, codegen_version) != nullptr)
    {
        results +=  buffer;
    }
    pclose(codegen_version);

    boost::algorithm::trim(results);
    return results;
}

const char* ChasteBuildInfo::GetXsdVersion()
{
    return "4.0.0";
}
```

Now because chaste_codegen is used in dynamic loading at runtime there could be a good reason for the above? But it does seem to cause a bottleneck at times so we should see if it can be optimised if it needs to be runtime.

mirams:
@jmpf @jonc125 @MauriceHendrix thoughts welcome. This is to do with issue #119

MauriceHendrix:
> Whereas other executables write their versions into the build-time-generated `Version.cpp`, chaste_codegen doesn't and does some weird looking file opening to read it. Think about whether we can put in the right info at build time.
> 
I think you're going to struggle to put the run-time codegen version number in the C++, since it's a python package that you install seperately (although it's done automatically during the build process). It's perfectly possible to switch to a different version after the build. 

If you don't mind the chance of that descrepancy, then it's probably possible.

fcooper8472:
I think you could technically switch other things, too. E.g. the version of XSD is hardcoded too:

```cpp
const char* ChasteBuildInfo::GetXsdVersion()
{
    return "${xsd_version}";
}
```
which is a fixed string after the CMake configure step. But it's exceedingly unlikely that anyone would replace the XSD executable between configuring and running Chaste.

Is there a realistic chance that someone would go into the virtual environment that's created at configure-time and change/update codegen without re-configuring?

My hunch is it ought to be safe to get CMake to fill in that string.

[...]

PR #123, opened 2023-06-26
Fix codegen version overhead

fcooper8472:
One possible way to approach #122 

This PR also tides `Version_cmake.cpp.in` and `Version.cpp.in`, consolidating into a single file.


mirams (approved):
Maybe worth mentioning the Version.cpp.in now relates to cmake not scons when we write up the release notes for removing scons.

PR #128, opened 2023-06-28
Add workflow to regenerate and upload crypt archive

fcooper8472:
Work in progress!

### DIFF
--- a/.github/workflows/regenerate-arhive.yml
+++ b/.github/workflows/regenerate-arhive.yml
@@ -2,9 +2,6 @@ name: Regenerate archive
 
 on:
   workflow_dispatch:
-  push:
-    branches:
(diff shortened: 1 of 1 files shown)


mirams:
You are a really useful engine Fergus (I have been watching Thomas the Tank Engine)

fcooper8472:
> You are a really useful engine Fergus (I have been watching Thomas the Tank Engine)

:steam_locomotive:

fcooper8472:
I'm being very naughty and merging this without review as the workflow (with a workflow dispatch trigger) needs to exist on the default branch in order to check that it works.

jmosborne:
If this is doing what I think its doing you're my new favourite person!

Issue #146, opened 2023-07-20
Investigate new Intel build failures


fcooper8472:
E.g. https://github.com/Chaste/Chaste/actions/runs/5609769767/jobs/10263908241?pr=145

```
[ 11%] Building CXX object global/test/CMakeFiles/TestDebugParallel.dir/TestDebugParallel.cpp.o
In file included from /home/runner/work/Chaste/Chaste/chaste-build-dir/global/test/TestArchivingHelperClassesParallel.cpp:21:
/home/runner/work/Chaste/Chaste/global/test/TestArchivingHelperClasses.hpp:68:9: error: exception of type 'Exception &' will be caught by earlier handler [-Werror,-Wexceptions]
        TS_ASSERT_THROWS_THIS(ArchiveLocationInfo::GetArchiveDirectory(),
        ^
/home/runner/work/Chaste/Chaste/chaste-build-dir/cxxtest/cxxtest/TestSuite.h:429:39: note: expanded from macro 'TS_ASSERT_THROWS_THIS'
#   define TS_ASSERT_THROWS_THIS(e,y) CHASTE_ASSERT_EXCEPTION_MESSAGE(__FILE__,__LINE__,e,CheckShortMessage,y)
                                      ^
/home/runner/work/Chaste/Chaste/chaste-build-dir/cxxtest/cxxtest/TestSuite.h:423:56: note: expanded from macro 'CHASTE_ASSERT_EXCEPTION_MESSAGE'
            _TS_CATCH_TYPE( (const Exception& _err), { TSM_ASSERT(_err.method(y), _err.method(y) == ""); _ts_threw_expected = true; } ) \
                                                       ^
/home/runner/work/Chaste/Chaste/chaste-build-dir/cxxtest/cxxtest/TestSuite.h:[249](https://github.com/Chaste/Chaste/actions/runs/5609769767/jobs/10263908241?pr=145#step:8:250):28: note: expanded from macro 'TSM_ASSERT'
#   define TSM_ASSERT(m,e) _TSM_ASSERT(__FILE__,__LINE__,m,e)
                           ^
note: (skipping 2 expansions in backtrace; use -fmacro-backtrace-limit=0 to see all)
/home/runner/work/Chaste/Chaste/chaste-build-dir/cxxtest/cxxtest/TestSuite.h:209:32: note: expanded from macro '__TS_CATCH'
#       define __TS_CATCH(f,l) __TSM_CATCH(f,l,"Unhandled exception")
                               ^
[...]

PR #147, opened 2023-07-20
#146 Remove explicit catch of Chaste Exception

fcooper8472:
See #146


fcooper8472:
I also merged the branch with new deb locations in so that the test suites actually run!

MILeach (approved):
All looks good to me and I think the exception handling change makes sense.

PR #152, opened 2023-08-09
Support VTK 9.x

### LINKED ISSUE #36: Support VTK 9.x
This issue continues the discussion for legacy trac ticket 3088:
___

[fcooper8472](https://github.com/orgs/Chaste/people/fcooper8472) created the following ticket on 2022-02-02 at 13:26:24, it is owned by [fcooper8472](https://github.com/orgs/Chaste/people/fcooper8472)

Changes can be seen here: https://www.kitware.com/vtk-9-0-0-available-for-download/

This currently prevents Chaste working (easily) on macOS as Homebrew gives v9.
___

Comment by [jmpf](https://github.com/orgs/Chaste/people/jmpf) on 2022-08-01 at 11:05:05

VTK9 was added to the rotation in February:

- [2d47416](https://github.com/Chaste/Chaste/commit/2d47416)
- [234186e](https://chaste.cs.ox.ac.uk/trac/changeset/234186e/infrastructure-scripts)

but it has yet to compile successfully: https://chaste.cs.ox.ac.uk/buildbot/builders/Portability%206_Sat?numbuilds=30
___

Comment by [jmpf](https://github.com/orgs/Chaste/people/jmpf) on 2022-08-04 at 13:33:48

Core components that need VTK are building fine (`global, mesh,` etc.).

The `cell_based` component fails when including `VtkMeshReader.hpp`. This is probably because it doesn't have the correct include path, but I can't work out why that may be.
___

Comment by [jmpf](https://github.com/orgs/Chaste/people/jmpf) on 2022-10-24 at 15:17:59

Replying to jmpf@…:

> Core components that need VTK are building fine (`global, mesh,` etc.).
> 
> The `cell_based` component fails when including `VtkMeshReader.hpp`. This is probably because it doesn't have the correct include path, but I can't work out why that may be.

Note that this is a CMake configuration problem and is to do with the way CMake now detects libraries and include paths.

fcooper8472:
See #36 

This approach creates a CMake interface library with no source files which our dependencies are linked against.

This interface library is then linked against chaste libraries and executables.

This allows us to use the modern [`target_link_libraries` ](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) method of ingesting dependencies, whereas we had previously been relying on populating lists of libraries and include directories, which is a much older CMake pattern.

VTK 9 is the first dependency that is no longer compatible with this old way of doing things, but part of the work on modernising our CMake infrastructure will be to migrate all our dependencies over to this new way of doing things.


jmpf (approved):
Changes look good.

Also tested on 22.04 (Jammy) with stock VTK 7.1.1 (vtk7-dev).
Had problems installing vtk9-dev without breaking multiple parts of my current Chaste development environment, but that's okay because CLI show it works on a fresh install.

MILeach (approved):

PR #153, opened 2023-08-15
Return Shared Pointers from Mesh Generators

### LINKED ISSUE #35: Use New Chaste Mesh Generators
**Description**
The mesh generators in [PyChaste/src/mesh](https://github.com/Chaste/PyChaste/tree/develop/src/mesh) are customised for PyChaste from the originals in [Chaste/mesh/src/vertex](https://github.com/Chaste/Chaste/tree/develop/mesh/src/vertex) and [Chaste/cell_based/src/mesh](https://github.com/Chaste/Chaste/tree/develop/cell_based/src/mesh):
* HoneycombMeshGenerator -> SharedHoneycombMeshGenerator
* CylindricalHoneycombMeshGenerator -> SharedCylindricalHoneycombMeshGenerator
* ToroidalHoneycombMeshGenerator -> SharedToroidalHoneycombMeshGenerator
* HoneycombVertexMeshGenerator -> SharedHoneycombVertexMeshGenerator
* CylindricalHoneycombVertexMeshGenerator -> SharedCylindricalHoneycombVertexMeshGenerator
* ToroidalHoneycombVertexMeshGenerator -> SharedToroidalHoneycombVertexMeshGenerator
* VoronoiVertexMeshGenerator -> SharedVoronoiVertexMeshGenerator
* PottsMeshGenerator -> SharedPottsMeshGenerator

The customised mesh generators are modified to return shared pointers instead of the original raw pointers. For example:
```diff
--- a/CylindricalHoneycombVertexMeshGenerator.hpp
+++ b/SharedCylindricalHoneycombVertexMeshGenerator.hpp

-    MutableVertexMesh<2,2>* GetMesh();
+    boost::shared_ptr<MutableVertexMesh<2,2> > GetMesh();

-    Cylindrical2dVertexMesh* GetCylindricalMesh();
+    boost::shared_ptr<Cylindrical2dVertexMesh> GetCylindricalMesh();
```

It would be cleaner to have this implemented in upstream Chaste to avoid the workaround.

**Related Issue**
* https://github.com/Chaste/Chaste/issues/151

### LINKED ISSUE #151: Return Shared Pointers from Mesh Generators
**Description**

* Currently, mesh generators are manually recreated in PyChaste and modified to return shared pointers instead of raw pointers. For example, `CylindricalHoneycombVertexMeshGenerator` is replaced by `SharedCylindricalHoneycombVertexMeshGenerator` in PyChaste:
```diff
--- a/CylindricalHoneycombVertexMeshGenerator.hpp
+++ b/SharedCylindricalHoneycombVertexMeshGenerator.hpp

-    MutableVertexMesh<2,2>* GetMesh();
+    boost::shared_ptr<MutableVertexMesh<2,2> > GetMesh();

-    Cylindrical2dVertexMesh* GetCylindricalMesh();
+    boost::shared_ptr<Cylindrical2dVertexMesh> GetCylindricalMesh();
```

* This is because the tight coupling between mesh generators and the meshes they generate in Chaste can lead to memory errors downstream. For example:
```python
generator = chaste.mesh.VoronoiVertexMeshGenerator(6, 6, 1)
voronoi_mesh = generator.GetMesh()
...
generator = chaste.mesh.PottsMeshGenerator2(50, 8, 4, 50, 8, 4)
potts_mesh = generator.GetMesh()
...
# The first generator object gets garbage collected, along with voronoi_mesh
...
voronoi_mesh.SetDistanceForT3SwapChecking(1.0) # Error!
```

* It would be more convenient to have this implemented directly in Chaste. Making use of `shared_ptr` would also support #139.

**Related Issues**
* #139 
* https://github.com/Chaste/PyChaste/issues/35

**Related PRs**
* https://github.com/Chaste/Chaste/pull/153
* https://github.com/Chaste/CellBasedComparison2017/pull/1

kwabenantim:
**Summary**

* Supports #151 
* List of changes: https://github.com/Chaste/Chaste/issues/151#issuecomment-1684936603

**Related PRs**

* https://github.com/Chaste/CellBasedComparison2017/pull/1


cell_based/src/mesh/PottsMeshGenerator.hpp
```diff
@@ -91,10 +92,12 @@ class PottsMeshGenerator
                        bool isPeriodicInZ = false);
 
     /**
-     * Destructor - deletes the mesh object and pointer.
```
mirams, line 94: @fcooper8472 might suggest these should use the `= default` rather than empty user-defined destructors (I think this can occasionally even save the compiler from making a destructor if there is no work to do, but haven't fully got my head around it!)

crypt/test/simulation/TestCryptSimulationAnother2dNightly.hpp
```diff
@@ -180,13 +180,13 @@ class TestCryptSimulationAnother2dNightly : public AbstractCellBasedWithTimingsT
         //double crypt_length = num_cells_depth - 1.0;
 
         HoneycombMeshGenerator generator(num_cells_width, num_cells_depth, 0);
-        MutableMesh<2,2>* p_mesh = generator.GetMesh();
+        boost::shared_ptr<MutableMesh<2,2> > p_mesh = generator.GetMesh();
         std::vector<unsigned> location_indices = generator.GetCellLocationIndices();
 
         // Set up cells
         std::vector<CellPtr> cells;
         CryptCellsGenerator<FixedG1GenerationalCellCycleModel> cells_generator;
```
mirams, line 188: Maybe a new ticket to get the CellsGenerators to take a `shared_ptr` instead of a raw pointer?

mirams (approved):
Looks OK, maybe a new ticket to get `CellsGenerator.Generate(...)` to take shared rather than raw pointer?

cell_based/src/mesh/PottsMeshGenerator.hpp
kwabenantim: Thanks @mirams. That seems like good practice according to the [core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all):
>**Example, good** When a destructor needs to be declared just to make it virtual, it can be defined as defaulted.

crypt/test/simulation/TestCryptSimulationAnother2dNightly.hpp
kwabenantim: Thanks @mirams. I can create a new ticket to do this as well. I've left it this way because the [core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f7-for-general-use-take-t-or-t-arguments-rather-than-smart-pointers) recommend taking non-smart pointer arguments for functions in general (unless they are going to take partial responsibility for managing the object or going to pass it forward to another function that will), and `CryptCellsGenerator::Generate()` only seems to be querying some properties of the mesh at the moment.

PR #159, opened 2023-09-14
Add extensions and tidy .devcontainer

bdevans:


bdevans:
@fcooper8472 I think you can ignore the failures on this - in an old habit, I initially pushed the changes to a personal fork of the main repo and created the PR from there meaning that it doesn't have the necessary infrastructure to run all the tests. 

I tried to rename the `test_summary` folder to `output` here too but it looks like I may have made a mistake as the old folder appears to have been deleted without committing the new one. If so, can you see how to fix that?

fcooper8472 (approved):

bdevans:
> I tried to rename the `test_summary` folder to `output` here too but it looks like I may have made a mistake as the old folder appears to have been deleted without committing the new one. If so, can you see how to fix that?

I believe this is now fixed.

PR #178, opened 2023-09-27
Add summary, and prioritise failing tests

fcooper8472:
This update will add a summary to the index page, and prioritise the failing tests to make the summaries easier to navigate.

The summary info will be parsed to shade the index table rows as green orange or red, making it easier to see at a glance which memory testing runs are failing.

Example of the new index page after this update:

image


jmpf (approved):
Cool.  Useful output.

Issue #185, opened 2023-11-06
Migrate how-to generation to GitHub


fcooper8472:
There is an index of "how-tos" here:
https://chaste.cs.ox.ac.uk/trac/wiki/ChasteGuides/HowTo

These are generated from specially-formatted comments in the source code of test files by a Python script in the infrastructure-scripts repository:

```
GenerateHowTo.py
```

It has been being run, as the wiki page was last updated in June, but it's unclear how. It's referenced in

```
run-doxygen.py
```

so it's possible it was generated with some Doxygen output. But, both scripts are Python 2 syntax, and we've not recently run anything with Python 2...

In any case, the best route forward is to re-write this as a new Python script that will generate a markdown file with the same information, and update a page on the github pages site, if necessary.

This is underway and this issue will be updated with progress.

fcooper8472:
This is underway, with a new script in the `python/infra` directory:

e62d6653bf8909949fda3701d3e29fcfa9885bb5

fcooper8472:
Not yet public on the menus, but this is now available here:
https://chaste.github.io/docs/how-tos/

@mirams does this look OK?

The only bug so far is I forgot the L in the line specification of the GitHub URL, so the links don't resolve to the correct line. But that will be fixed in an upcoming commit.

mirams:
Yeah, functionally looking good. 

Presentation-wise a few improvements probably.

First one is just about the whole website I guess. Trac used to use horizontal space well (wiki could include line breaks if we wanted them). The text on the trac website isn't much smaller than the new website, but look how much more scrolling you need to do now.
<kbd><img width="1716" alt="image" src="https://github.com/Chaste/Chaste/assets/858776/7030e009-a61d-48a8-b6d2-d9f27957bed4"></kbd>

new one (when even the quick link side menu doesn't fit on the whole screen!):
<kbd><img width="1724" alt="image" src="https://github.com/Chaste/Chaste/assets/858776/189f613f-0180-42c6-8853-ad6e93278365"></kbd>

Other things are a couple of links can be added:
<kbd><img width="736" alt="image" border=1 src="https://github.com/Chaste/Chaste/assets/858776/7aa1baf1-8a03-47b6-8811-39f6e8294725"></kbd>
this bullet point had a link to doxygen class documentation and also wiki page on how to do Cellml to code. I guess these now could be just later updates to the comments to use new markdown link format if we don't want to do anything special.

[...]

PR #189, opened 2023-11-14
Automatically generate how-tos

fcooper8472:
See #185 

This PR adds a generation python script, and a github actions workflow that will automatically update the website.


mirams (dismissed):

fcooper8472:
@mirams this is now ready for review, and has generated the following how-tos page:
https://chaste.github.io/docs/how-tos/

#185 has been updated with the details.

.github/workflows/update-how-tos.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: |
+          python GenerateHowTos.py
+        working-directory: python/infra
+
+      - name: Upload archive files
+        uses: actions/upload-artifact@v3
+        with:
+          name: how-to-markdown
+          path: |
+            chaste_howto_webpage.md
```
MILeach, line 33: Just checking this filename/path doesn't want another underscore in `howto`

MILeach (approved):
Looks good to me!

.github/workflows/update-how-tos.yml
fcooper8472: it's just a temp file uploaded as an artifact between jobs - it's actually copied to _index.md when it becomes part of the website.

mirams (approved):

mirams:
Looks a lot tidier Fergus, thanks. I also very much like the auto-highlighting right hand panel

fcooper8472:
The memory testing failure was only failing to generate the index file due to missing python... this seems like a spurious error, but one that should perhaps be addressed. I'll open a new issue for that.

In the meantime I'll merge this.

PR #204, opened 2024-01-10
Update LICENSE copyright

mirams:


jmpf (approved):
Would be good to regularly check for copyright notices

mirams:
I think we probably do it in the infrastructure tests don't we? (we used to). But it won't fail until someone manually tells it what the current notice should look like at the moment.

PR #206, opened 2024-01-19
Docker-release-ci

bdevans:
This adds a new GitHub action in readiness for the next release to automate the building of the corresponding `chaste/release` image. 
It also renames the existing `chaste/develop` image builder action and adds `build-args` to accomodate some planned changes to the `chaste/base` image.


kwabenantim (approved):
The docker release workflow looks fine to me @bdevans.

PR #207, opened 2024-01-23
Fix error with GenerateUserTutorials.py

mirams:
On python 3.8.10 it was saying
`TypeError: 'type' object is not subscriptable` this seems to be recommended fix.


fcooper8472:
Ah, this is because they changed type hinting in Python 3.9 to allow this syntax. Python 3.8 is still supported for a few more months, so I guess we should fix the script!

Your proposed fix will work, but prevents the interpreter itself from checking the types.

The better fix would be:

```python
from typing import List, Tuple
```

And then change to `List[Path]`, `List[str]` and `Tuple[str, List[str]]`.

mirams:
OK, seems to work!

fcooper8472 (approved):

Issue #224, opened 2024-02-22
Improvements to tutorials


mirams:
Now the tutorials are on the new website, following #193 , time for a spring clean. 

A few improvements we noted during the migration:
## Top Priority
- [x] Specify where test output goes, both in general & in docker (@bdevans). Could either be on the first tutorial https://chaste.github.io/docs/user-tutorials/writingtests/  or mentioned in big letters on the tutorial index page (I probably prefer the latter).

## Nice to have
- [ ] Move java visualiser code to separate repo (code lives in the anim folder). Keep `anim` folder but replace code with a readme pointing to its new location.
- [ ] Remove references to java visualiser from tutorials / docs.
- [ ] Update paraview tutorial to make it a lot more novice-user friendly.

bdevans:
So we recently renamed the defunct `test_summary` folder in the main repo to `output` so that there is a known, useable folder in the repo that VSCode can automatically use as a mount when someone clones the repo and opens it in a devcontainer. For uniformity, I also changed the environment variable `CHASTE_TEST_OUTPUT="${CHASTE_DIR}/output"` in the docker images to match (i.e. by default it will be `/home/chaste/output`). This way the path is shorter and more general too as people could produce other types of output there (and could if they wish, still have a subfolder for test outputs etc.).

mirams:
OK, good, in that case it might be good enough to make the tutorials refer to `CHASTE_TEST_OUTPUT` and not need any docker specialisation.

bdevans:
I agree - a level of abstraction by using an environment variable makes it all a lot clearer, more uniform and much more flexible.

PR #226, opened 2024-02-22
Make tests point to `CHASTE_TEST_OUTPUT`

mirams:
to be friendlier for Docker users. Work on #224


mirams:
cough @fcooper8472 !

fcooper8472:
Discussion required (most relevant probably to @mirams and @jmpf):

At the moment the `FileFinder.hpp` defines the following struct:

```cpp
struct RelativeTo
{
    /**
     * What things a path can be relative to.
     */
    enum Value
    {
        CWD, /**< The current working directory */
        ChasteTestOutput, /**< The CHASTE_TEST_OUTPUT folder */
        ChasteSourceRoot, /**< The root of the Chaste source tree */
        ChasteBuildRoot, /**< The root of the Chaste build tree */
        Absolute, /**< This is an absolute path */
        AbsoluteOrCwd /**< If it starts with a / then it's absolute, otherwise interpret relative to CWD */
    };
};
```

The `ChasteSourceRoot` and `ChasteBuildRoot` are handled by functions defined in `ChasteBuildInfo_cmake.cpp.in`, which is filled in by CMake at configure time.

The `ChasteTestOutput` is more complicated, and uses logic in `OutputFileHandler::GetChasteTestOutputDirectory()` that uses `getenv` to get the `CHASTE_TEST_OUTPUT` environment variable.

My suggestion is we simplify this logic. I can't see an important use-case for changing the test output directory more often than could be achieved by re-configuring Chaste with CMake. This would allow us to bake the output directory into the `ChasteBuildInfo_cmake.cpp.in` and CMake would determine whether (at configure time) to set that as the value `CHASTE_TEST_OUTPUT` or, else, a new subdirectory of the build directory.

Does this sound OK?

mirams:
Should be OK I guess. But it does mean that it becomes impossible to distribute an executable that writes to a settable `CHASTE_TEST_OUTPUT` (e.g. my main use-case worry would be compiling on a supercomputer login node, and wanting to run tests elsewhere on other nodes, but I guess the file structures are usually shared ?? ). If the only benefit is simplifying logic in `OutputFileHandler` I'd probably keep it as it is, or is there a benefit in this move to put the default location under `ChasteBuildRoot`?

fcooper8472 (approved):

Issue #232, opened 2024-02-27
Utilise larger VMs available by default on GitHub hosted runners


fcooper8472:
GitHub hosted runners have become bigger for public repos:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

The default Linux runners have 4 CPU cores, 16GB RAM and 14GB SSD storage.

This means we can:

- [x] Build and run tests with 4 cores by default
- [ ] Move the np3 runner back into the cloud
- [x] Possibly(?) move the memtest runner back into the cloud

fcooper8472:
> Build and run tests with 4 cores by default

Done, and has been for a long time

> Move the np3 runner back into the cloud

Can't be done, because it's still only 2 physical cores (4 threads), and we need at least 3 physical cores. Still running self-hosted.

> Possibly(?) move the memtest runner back into the cloud

Done, and has been for a long time

PR #233, opened 2024-02-27
Utilise larger GitHub hosted runners

fcooper8472:
See #232

### DIFF
--- a/.github/workflows/docker-develop-image.yml
+++ b/.github/workflows/docker-develop-image.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted-docker
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
--- a/.github/workflows/docker-release-image.yml
+++ b/.github/workflows/docker-release-image.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted-docker
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -22,7 +22,7 @@ jobs:
         RUNNER_OFF: 1
       volumes:
         - runner_volume:/home/runner
-      options: --user 0 --cpus 2
+      options: --user 0 --cpus 4
       
     env:
       CC: gcc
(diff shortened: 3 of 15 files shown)


.github/workflows/docker-develop-image.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted-docker
```
fcooper8472, line 12: @bdevans does this look appropriate? There is a dedicated runner tagged with this label, but are there other changes likely to be needed? I'm not sure we can currently test this workflow without merging.

.github/workflows/docker-release-image.yml
```diff
@@ -13,7 +13,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted-docker
```
fcooper8472, line 16: @bdevans does this look appropriate? There is a dedicated runner tagged with this label, but are there other changes likely to be needed? I'm not sure we can currently test this workflow without merging.

.github/workflows/docker-develop-image.yml
bdevans: Yes, assuming you have tagged the machine with `self-hosted-docker`(before it was just `docker`). 
Thanks, once that's merged it should start working again...! 🤞

fcooper8472:
I've put in a temporary fix for coverage in the random field generation code; see #234 for a longer-term solution.

.github/workflows/docker-develop-image.yml
bdevans: Actually, if you tagged it as `docker` then this would mean we could rerun the original job to build the `chaste/release` image and test that it works. The tag could then be changed to the more descriptive `self-hosted-docker` afterwards.

fcooper8472:
f23ec4a2a47032e3db5393e291a343198aea5d42 adds some statistically appropriate tests for random field generation. This is enough to sensibly fix the coverage error, but further work is needed. See #234.

fcooper8472:
Pending green ticks, this should be ready for review.

@bdevans or @kwabenantim, could you review with a mind to the GitHub runner changes?

@AlexFletcher @mirams or @jmpf, could one of you take a quick look at the testing changes to random fields? They're related to #234, but everything is blocked until we have a green tick on coverage, so the minimum work on that has been added here.

bdevans (approved):
👍

kwabenantim (commented):
The workflow changes look good to me.

fcooper8472:
The coverage fix does not need to be mentioned in changes since last release, as it just fixes two lines that were uncovered at the point of the last release.

fcooper8472:
@jmpf approves.

Issue #133, opened 2023-07-04
Move Svn (Trac/data/re-directions) onto new server


jmpf:
Subtask of #39

jmpf:
I installed trac (which pulled in apache2 and libsvn ) on 23 May 2023.  Apache working and tested, but trac needed some thinking about.

jmpf:
**Cloning SVN data**
```
# As svn@chaste (which has permission to use mount /var/www on large drive)
 svnadmin dump /var/svn/chaste/ > /var/www/chaste.svn_dump
```
(That takes about an hour)

```
# On newchaste.cs
sudo apt install subversion
sudo mkdir /home/svn
sudo useradd svn -g www-data -d /home/svn

# As svn@newchaste
scp chaste@chaste:/var/www/chaste.svn_dump .
```
**Failed** The server has 100Gb of root SSD (total).  This file is about 53 Gb

jmpf:
Updated logical volume from 100Gb to 500Gb.

```
sudo lvresize --size 500G ubuntu-vg/ubuntu-lv
sudo resize2fs /dev/mapper/ubuntu--vg-ubuntu--lv 
```

jmpf:
Set up the Subversion server on the new machine:
```
sudo mkdir /var/svn
sudo chown svn /var/svn 
sudo su - svn
svnadmin create /var/svn/chaste
svnadmin load /var/svn/chaste < /home/github/chaste.svn_dump
```

jmpf:
```
# As svn@chaste:
trac-admin /opt/tracdb/chaste/ hotcopy chaste-hotcopy
```
```
# On newchaste
mkdir /opt/tracdb
sudo mkdir /opt/tracdb
sftp -r chaste@chaste:/home/svn/chaste-hotcopy .

# sudo su - svn
cp -r /home/github/chaste-hotcopy/ /opt/tracdb/
# Don't do the next line yet.  We might need to initialise first...
mv /opt/tracdb/chaste-hotcopy/ /opt/tracdb/chaste
```
Now need to work out what it takes to get trac-admin to work...

jmpf:
Read https://trac.edgewall.org/ticket/13404 and so I patched the Ubuntu Trac distribution
```
--- /usr/lib/python3/dist-packages/trac/util/presentation.py.0  2023-08-11 14:59:47.877759479 +0000
[...]

PR #237, opened 2024-03-07
Create redirection_recommendation template

fcooper8472:
Template issue for a user to recommend a new redirection from chaste.cs -> chaste.github.io

This is related to #133 : when trac goes offline we need to ensure users are redirected appropriately.

We will be able to get most links that are in the wild (e.g. paper tutorials and the downloads page) working appropriately, but we will also introduce a catch-all page (see https://github.com/Chaste/Chaste.github.io/issues/83) for any redirects we haven't anticipated.

That page can link to this new issue template so that if a user comes across a link in the wild that ought to redirect to a specific place, they can make us aware of it.


kwabenantim (approved):
This seems like a good idea to me. If it's for users to report links that have stopped working for them, maybe a gentle intro along the lines of "we've moved Chaste docs from chaste.cs.ox.ac.uk/trac to chaste.github.io" might be useful. Also, a link to the catch-all page would help as well.

fcooper8472:
> This seems like a good idea to me. If it's for users to report links that have stopped working for them, maybe a gentle intro along the lines of "we've moved Chaste docs from chaste.cs.ox.ac.uk/trac to chaste.github.io" might be useful. Also, a link to the catch-all page would help as well.

Good point - I had assumed people would only find this from the catch-all page, but of course someone could easily find this directly through the Chaste repo. I'll add something appropriate and then merge.

fcooper8472:
This is no longer needed due to @jmpf's work: we have pretty robust redirections set up.

PR #242, opened 2024-03-12
Replace depreacted since CMake 3.12 FindPythonInterp with FindPython3

### LINKED ISSUE #236: CMake >= 3.27 FindPythonInterp removal
CMake 3.12 deprecated the `FindPythonInterp` and `FindPythonLibs` modules, and 3.27 subsequently removed them when [CMP0148](https://cmake.org/cmake/help/latest/policy/CMP0148.html) is set to NEW.

Configuring Chaste with CMake >= 3.27 currently (b98110e9d82a10460375695d02b4040047da1734) emits the following dev-warning, which will be visible to users:

```
CMake Warning (dev) at /home/ptheywood/code/chaste/chaste/cmake/Modules/ChasteGenerateVersionAndBuildInfo.cmake:89 (find_package):
  Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
  are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

This warning is for project developers.  Use -Wno-dev to suppress it.
```

`find_package(PythonInterp)` is called at 4 locations within Chaste/Chaste:

+ `CMakeLists.txt`
+ `cmake/Config/ChasteConfig.cmake.in`
+ `cmake/modules/ChasteGenerateVersionAndBuildInfo.cmake`
+ `CellMLToSharedLibraryConverter::ConvertCellmlToSo` in `heart/srtc/odes/CellMLToShareLibraryConverter.cpp`.

The simplest replacement should be to use `find_package(Python)` or `find_package(Python3)`, optionally with `COMPONENTS Interpreter` (but that's the default behaviour) instead, which has been available since CMake 3.12.

Additionally the cmake variables `PYTHON_EXECUTABLE` and `PYTHON_VERSION_STRING` defined by [`FindPythonInterp`](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html) and used by chaste would need replacing with the equivalent value from [`findPython`](https://cmake.org/cmake/help/latest/module/FindPython.html)/[`findPython3`](https://cmake.org/cmake/help/latest/module/FindPython3.html) (`Python_EXECUTABLE`/`Python3_EXECUTABLE` and `Python_VERSION`/`Python3_VERSION`).

The [dependency versions documentation](https://chaste.github.io/docs/installguides/dependency-versions/) suggests < 3.16.3 is not supported,
however, there are `cmake_minimum_required` statements with a range of versions within Chaste, including a relevant location:

+ `3.16.3` (`CMakeLists.txt`, heart/srtc/odes/CellMLToShareLibraryConverter.cpp)
+ `2.8.10` (`cmake/CMakeLists.txt`)
+ `3.10.2` (`cmake/Config/ChasteConfig.cmake.in`)
+ `2.8/2.8.10` in `third_party_libs`, but this is probably unaffected

Some of these (`cmake/Config/ChasteConfig.cmake.in` at least) would either need updating to be >= `3.12` (or `3.16.3` for parity), or conditionally use the deprecated `FindPythonInterp` based on CMake version. Given the difference in defined CMake Variables, only using the modern approach seems like the preferred option if 3.16.3 is a strict minimum CMake version.

This is encountered on ubuntu-22.05 build-and-test CI workflows during the configure phase ([example](https://github.com/Chaste/Chaste/actions/runs/8186262983/job/22384311184)), but cmake dev and deprecated warnings are not promoted to errors. This could optionally be enabled by passing `-Werror=dev` to CMake.

ptheywood:
Replace depreacted since CMake 3.12 `FindPythonInterp` with `FindPython3`, which emits author warnings from CMake 3.27 and will be removed from a future CMake release.

Also updates cmake_minimum_required to be consistent to ensure FindPython3 is avialable, excluding 3rd party deps

Closes #236

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,9 +273,9 @@ endif ()
 ################################
 ####  Find Python
 ################################
-find_package(PythonInterp 3.5)
-if (PYTHON_VERSION_STRING VERSION_LESS 3.5.0)
-    message(FATAL_ERROR "Python 3.5 or higher is required; found v${PYTHON_VERSION_STRING} instead. Specify interpreter with -DPYTHON_EXECUTABLE=/path/to/python3")
+find_package(Python3 3.5)
+if (Python3_VERSION VERSION_LESS 3.5.0)
+    message(FATAL_ERROR "Python 3.5 or higher is required; found v${Python3_VERSION} instead. Specify interpreter with -DPython_EXECUTABLE=/path/to/python3")
 endif ()
 
 # Install chaste_codegen in virtual environment
@@ -284,14 +284,14 @@ set(codegen_python3_venv ${PROJECT_BINARY_DIR}/codegen_python3_venv/bin)
 if (NOT EXISTS ${codegen_python3_venv})
     message (STATUS "Creating virtual environment for chaste_codegen in ${PROJECT_BINARY_DIR}/codegen_python3_venv")
     execute_process(
-        COMMAND ${PYTHON_EXECUTABLE} -m venv codegen_python3_venv
+        COMMAND ${Python3_EXECUTABLE} -m venv codegen_python3_venv
         WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
     )
 endif ()
 
 # Update pip setuptools and wheel
 message (STATUS "Updating pip setuptools and wheel")
-if (${PYTHON_VERSION_STRING} VERSION_LESS 3.6.0)
+if (${Python3_VERSION} VERSION_LESS 3.6.0)
     execute_process(COMMAND ${codegen_python3_venv}/pip install --no-cache-dir "pip>=20.0, <21.0")
     execute_process(COMMAND ${codegen_python3_venv}/pip install --no-cache-dir --upgrade setuptools wheel)
 else()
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -273,9 +273,9 @@ endif ()
 ################################
 ####  Find Python
 ################################
-find_package(PythonInterp 3.5)
-if (PYTHON_VERSION_STRING VERSION_LESS 3.5.0)
-    message(FATAL_ERROR "Python 3.5 or higher is required; found v${PYTHON_VERSION_STRING} instead. Specify interpreter with -DPYTHON_EXECUTABLE=/path/to/python3")
+find_package(Python3 3.5)
```
fcooper8472, line 276: We could probably be much more aggressive with this version number now. These changes need to work with Ubuntu 20.04 but nothing earlier, so think this should be 3.8.

Not relevant for this PR specifically, but @ptheywood what do you think? Could you open a new issue for us to decide an appropriate minimum Python version?

fcooper8472 (approved):

CMakeLists.txt
ptheywood: <= 3.7 is EOL since summer 2023, so it seems like the sensible current minimum version. 

I've opened #246

Issue #264, opened 2024-04-29
Ubuntu 24.04 PETSC citation update


fcooper8472:
The PETSc citation needs updating.

It's not clear to me how to find the appropriate citation, or where it should go, but it has been updated at least once since 3.15.

Perhaps @jmpf or @mirams could have a look?

mirams:
This looks as if it was the last time we had to do this:
https://github.com/Chaste/Chaste/commit/562005227f9c7160dc45e00b76e8fadc4cec19e2.

I think Fergus Cooper is something of an expert though: https://github.com/Chaste/Chaste/commit/c24e79522672256bab230223911d7d0e8571a8a7

(probably the Chaste default citation wants updating too whilst we're at it!)

fcooper8472:
Ha! Good find. I have no recollection of doing those.

I've just had a play around and I can't find where the updated citations are... but I imagine it's documented somewhere in PETSC.

fcooper8472:
Closed by #276

PR #276, opened 2024-05-28
Update PETSc citation for version with Ubuntu 24.04

fcooper8472:
See #264


fcooper8472:
@mirams / @jmpf because of the libexpat we can't yet get Ubuntu 22.04 testing into the rotation, but I can confirm that this change passes on Ubuntu 22.04.

mirams (approved):

Issue #193, opened 2023-11-22
Migrate tutorials to new website


fcooper8472:
This issue is to track any remaining problems with how tutorials are being displayed on the website:

https://chaste.github.io/docs/user-tutorials/

AlexFletcher:
I've just had a look through the cell-based user tutorials and have spotted the following issues - some of these go beyond syntax but I thought I'd record them here while they were on my mind:

[Cell Based Demo](https://chaste.github.io/docs/user-tutorials/cellbaseddemo)
- [x] Broken link: "[wiki:UserTutorials/CreatingAndUsingANewCellProperty]"
- [x] Formatting: in "using the MAKE_PTR macro as follows", format "MAKE_PTR" as code
- [x] Broken link: "[wiki:UserTutorials/VisualizingWithParaview]"

[Creating And Using A New Cell Based Simulation Modifier](https://chaste.github.io/docs/user-tutorials/creatingandusinganewcellbasedsimulationmodifier)
- [x] Formatting: in "in the CellData property", format "CellData" as code

[Creating And Using A New Cell Cycle Model](https://chaste.github.io/docs/user-tutorials/creatingandusinganewcellcyclemodel)
- [x] Formatting: in "the Boost shared_ptr smart pointer", format "shared_ptr" as code
- [x] Formatting: in "the NEVER_REACHED macro", format "NEVER_REACHED" as code
- [x] Formatting: in "except for except for CheckReadyToDivideAndPhaseIsUpdated", for consistency should we add brackets to emphasise that CheckReadyToDivideAndPhaseIsUpdated is a function, as is done elsewhere?
- [x] Broken link: "ChasteGuides/BoostSerialization"
- [x] Formatting: in "the concrete class MyCellCycleModel", format "MyCellCycleModel" as code

[Creating And Using A New Cell Killer](https://chaste.github.io/docs/user-tutorials/creatingandusinganewcellkiller/)
- [x] Formatting: in "the equation (‘‘x’’/20)^2^ + (‘‘y’’/10)^2^ < 1", remove the " and extra ^
- [x] Broken link: "[wiki:UserTutorials/CreatingAndUsingANewCellCycleModel]"
- [x] Formatting/typo: in "the macro MAKR_PTR_ARGS", format "MAKR_PTR_ARGS" as code and fix the typo in it

[Creating And Using A New Cell Mutation State](https://chaste.github.io/docs/user-tutorials/creatingandusinganewcellmutationstate)
[...]

PR #282, opened 2024-06-05
Pull request for adding simple command line based tutorial.

### LINKED ISSUE #267: Create user tutorial for cell-based simulation command line arguments and parameter sweeping
See https://github.com/Chaste/trac_archive/blob/6a7d720d0f6351f9b0d2c7b9b3464f4f265cfa41/attachment/ticket/2762/ExampleCellBasedExecutable.cpp, old tickets #2601 and #2762, and old wiki pages for inspiration; see also Boost program options and `CommandLineArguments`.

See also work done by @twinkarma as part of a previous BBSRC grant: https://github.com/twinkarma/chastesweep

BJackal:
Added two new tutorials for executing Chaste tests with command line arguments through bash scripts. 

The first is a very simple Hello World style example which takes in multiple command line arguments and outputs the sum. For now this has been placed in global as it seemed the most appropriate location.

Another example has been added for running a bash script with command line arguments to execute multiple vertex based simulations with the height and width of tissue based on command line parameters. This is a slightly expanded version of the vertex based simulation tutorial.


global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,37 @@
+#ifndef TESTCOMMANDLINEARGUMENTSTUTORIAL_HPP_
```
mirams, line 36: Needs the copyright notice (this is causing a [test to fail](https://github.com/Chaste/Chaste/actions/runs/9385768547/job/25844727324?pr=282)). Should be a matter of either copying and pasting from another test or running the script mentioned in the failure message.

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @file
+ *
+ * This is a tutorial for providing command line arguements to a test from a bash script.
+ * The example bash script also includes a for loop as an example of how such input could
+ * be used to run multiple tests with varying parameter inputs.
+ */
+
+ // NOTE: This test will not work if directlly executed from terminal due to requiring command line arguements.
```
mirams: Sounds as if it will make the test pack fail as well then? We can't really be having that, might need to put heads together on how to do a tutorial a different way...

It should be easy to do an "overwriting defaults" using e.g.
```c++
unsigned outp1 = 0;
if CommandLineArguments::Instance()->OptionExists("-opt1")
{
    outp1 = CommandLineArguments::Instance()->GetUnsignedCorrespondingToOption("-opt1");
}
```
and then we can put in comments saying "Please rerun the executable with arguments to see how the output changes, but note you will have to do this outside the `ctest` framework".

mirams:
```suggestion
 // NOTE: This test will not work if directly executed from terminal due to requiring command line arguments.
```

BJackal: I can think of one alternative way. The test could have some default conditions that are ran if command line arguments are not detected. Allowing the tutorial to simply run with those default initial conditions.

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+ // NOTE: This test will not work if directly executed from terminal due to requiring command line arguments.
+ // Investiagte the accompanying runcommandlinetutorial.sh bash script to see how this tutorial should be executed.
+class TestCommandLineArgumentsTutorial : public CxxTest::TestSuite
+{
+public:
+    void TestCommandLineTutorial()
+    {
+        // Here we utilise CommandLineArguments::Instance()->GetUnsignedCorrespondingToOption() to take in our command line arguements.
+        int outp1 = CommandLineArguments::Instance()->GetUnsignedCorrespondingToOption("-opt1");
```
mirams: variable is an `int` but that should return an `unsigned` (int)? Not sure all compilers will like that or whether we should suggest as best practice in a tutorial, I would replace with unsigned variables here probably

global/test/TestCommandLineArgumentsTutorial.hpp
BJackal: Corrected in https://github.com/Chaste/Chaste/pull/282/commits/3c90aa18a90573a59675f081acadf022734b3a30

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,72 @@
...
+ * The example bash script also includes a for loop as an example of how such input could
+ * be used to run multiple tests with varying parameter inputs.
+ */
+
+ // NOTE: This test will not work if directly executed from terminal due to requiring command line arguments.
+ // Investiagte the accompanying runcommandlinetutorial.sh bash script to see how this tutorial should be executed.
+class TestCommandLineArgumentsTutorial : public CxxTest::TestSuite
+{
+public:
+    void TestCommandLineTutorial()
```
mirams: Probably also want to show off some of the other nice methods in CommandLineArguments. e.g. getting `double`s and also getting vectors is a nice one with e.g.
```
./executable --my-vector-of-arguments 1.1 2.2 3.3 4.4
```
using 
```
std::vector<double> some_doubles = CommandLineArguments::Instance()->GetDoublesCorrespondingToOption("--my-vector-of-arguments");
```

global/test/TestCommandLineArgumentsTutorial.hpp
BJackal: This will similarly be changed for TestCommandLineArgumentsTutorial

cell_based/test/tutorial/TestCommandLineVertexMeshTutorial.hpp
```diff
@@ -0,0 +1,211 @@
+/*
```
mirams: I'm not really sure we need a whole vertex simulation here, it is bloating things a bit for no real benefit (there's no need for it to be a big complicated example, and nothing cell-based specific about passing in arguments). 

What if we just copy and paste the bash script into the `TestCommandLineArgumentsTutorial.hpp` to be rendered as a block of code on the website for readers to paste into a bash script? To e.g. just run the tutorial with different inputs in a loop?

mirams (changes requested):
A few suggestions, I think I would simplify quite a bit as there's nothing vertexey or cell based about wanting to pass in arguments, and would complicate the tutorial to show off more of the capabilities of the `CommandLineArguments` class!

cell_based/test/tutorial/TestCommandLineVertexMeshTutorial.hpp
BJackal: I can see this being very useful. Creating multiple tests within `TestCommandLineArgumentsTutorial.hpp` with a commented bash script for each. Then informing the user to copy and paste each variant of the bash script into their file for each test they wish to try out. Each test could then be showing how to execute with various methods, such as using vectors. As you mentioned this would look much cleaner on the website also.

BJackal:
1. Removed the vertex based simulation and its accompanying bash script.
2. Added a secondary test for the TestCommandLineArgumentsTutorial using a vector of doubles as the input.
3. Added a commented section to each test with the required bash script to run them, by default left the simpler test as the pre-existing bash script.
4. Changed int in the first test to unsigned as suggested.
5. The bash script to take in doubles requires awk (or a similar other tool) to allow for doubles to be input as arguments. 
6. Added a default case for each test if the correct corresponding command line argument is not detected to prevent the test failing.

Possible further additions I can think of now, it may be best to add some simple testing into the default cases or exceptions/assertions to ensure it is running as expected. Additionally , as the CommandLineArguments has the `GetBoolCorrespondingToOption` method should a test just showing this also be added ?

global/test/TestCommandLineArgumentsTutorial.hpp
BJackal: Edited as suggest

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+ // tutorial tests are intended to be executed with the accompanying runcommandlinetutorial.sh bash script.
+ // Investiagte the accompanying runcommandlinetutorial.sh bash script to see how this tutorial should be executed.
+class TestCommandLineArgumentsTutorial : public CxxTest::TestSuite
+{
+public:
+
+    /* The Default test should be ran with the following bash script starting from line 16.
+    runcommandlinetutorial.sh by default has this script already. However, if you have changed
+    the bash script for another text simply copy and paste between START and END back into the file
+    **START**
```
mirams:
```suggestion
    ```sh
```

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+      for ((i = 0; i <= N; i += 1)); do
+        for ((j = 1; j <= L; j += 1)); do
+          for ((k = 2; k <= M; k += 1)); do
+          # Each loop runs an instance of the TestCommandLineArgumentsTutorial with opt1,opt2 and opt3 taking on the 
+          # values of i,j and k resepctivley.      
+          ~/build/global/test/TestCommandLineArgumentsTutorial -opt1 $i -opt2 $j -opt3 $k &
+          done
+        done
+      done
+    **END**
```
mirams:
```suggestion
    ```
```

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+        
+        std::cout << "When we add "<< outp1 << " ,"<<  outp2 << " and "<< outp3 <<" we get " << sum <<"\n";
+    }
+    
+    /*
+      In addition to using unsigned integers as command line arguments we can also pass in both doubles and a vector of arguments.
+      This could be very useful for several reasons, such as having a large list of parameters one may wish to pass into a test. 
+
+      For this test change the runcommandlinetutorial.sh bash script from line 16 onwards as follow by copy and pasting between START and END.
+      **START**
```
mirams:
```suggestion
      ```sh
```

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+           jdouble=$(awk "BEGIN {printf \"%.2f\",$j/150}")
+           kdouble=$(awk "BEGIN {printf \"%.2f\",$k/180}")
+
+           # Each loop runs an instance of the TestCommandLineArgumentsTutorial with a vector
+           # containing the double corrected version of our varaibles.
+           ~/build/global/test/TestCommandLineArgumentsTutorial --my-vector-of-arguments $idouble $jdouble $kdouble &
+        done
+       done
+      done
+      **END**
```
mirams:
```suggestion
      ```
```

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+           // Again we need to setup our vector of doubles.
+           std::vector<double> vector_of_doubles{1.001, 2.002, 3.003};
+           double sum= 0.0;
+           // Again summing over our vector.
+           for(double i = 0; i < vector_of_doubles.size(); i++){
+            sum += vector_of_doubles[i]; 
+           }
+           // We will finally cout our result for the summed vector components.
+           // Here we have chosen to set our precision to 4 to ensure the correct number of significant figures.
+           std::cout << std::setprecision(4) << "When we add all variables in our vector together we get " << sum  <<"\n";
```
mirams:
```suggestion
           std::cout << std::setprecision(4) << "When we add all variables in our vector together we get " << sum  <<std::endl; 
```

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+           // First, we need to setup our vector based on the vector passed in as our command line arguement. 
+           std::vector<double> vector_of_doubles = CommandLineArguments::Instance()->GetDoublesCorrespondingToOption("--my-vector-of-arguments");
+           double sum = 0.0;
+           // Now summing over our vector.
+           for(double i = 0; i < vector_of_doubles.size(); i++){
+            sum += vector_of_doubles[i]; 
+           }
+           // We will finally cout our result for the summed vector components.
+           // Here we have chosen to set our precision to 4 to ensure the correct number of significant figures.
+           std::cout << std::setprecision(5) << "When we add "<< vector_of_doubles[0] << ", " << vector_of_doubles[1] << " and " << vector_of_doubles[2] << " together from our vector we get " << sum  <<"\n";
```
mirams:
```suggestion
           std::cout << std::setprecision(5) << "When we add "<< vector_of_doubles[0] << ", " << vector_of_doubles[1] << " and " << vector_of_doubles[2] << " together from our vector we get " << sum  <<std::endl; 
```

global/runcommandlinetutorial.sh
mirams, line 1: can we just remove this one if it is in the generated tutorial?

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+           }
+           // We will finally cout our result for the summed vector components.
+           // Here we have chosen to set our precision to 4 to ensure the correct number of significant figures.
+           std::cout << std::setprecision(4) << "When we add all variables in our vector together we get " << sum  <<"\n"; 
+           
+        }
+    }
+};
+
+#endif /*TESTHELLO_ECADTURNOVERMODEL_HPP_*/
```
mirams:
```suggestion
#endif /*TESTCOMMANDLINEARGUMENTSTUTORIAL_HPP_*/
```

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+           double sum = 0.0;
+           // Now summing over our vector.
+           for(double i = 0; i < vector_of_doubles.size(); i++){
+            sum += vector_of_doubles[i]; 
+           }
+           // We will finally cout our result for the summed vector components.
+           // Here we have chosen to set our precision to 4 to ensure the correct number of significant figures.
+           std::cout << std::setprecision(5) << "When we add "<< vector_of_doubles[0] << ", " << vector_of_doubles[1] << " and " << vector_of_doubles[2] << " together from our vector we get " << sum  <<"\n"; 
+
+        } else {
```
mirams: Can we just remove the else block, it is not doing anything here?

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+      L=3
+      M=4
+
+      # Here we set up a simple for loop over variables i,j and k based on the values of N,L and M.
+      for ((i = 0; i <= N; i += 1)); do
+        for ((j = 1; j <= L; j += 1)); do
+          for ((k = 2; k <= M; k += 1)); do
+          # Each loop runs an instance of the TestCommandLineArgumentsTutorial with opt1,opt2 and opt3 taking on the 
+          # values of i,j and k resepctivley.      
+          ~/build/global/test/TestCommandLineArgumentsTutorial -opt1 $i -opt2 $j -opt3 $k &
```
mirams: Path here is hardcoded, suggest altering instructions to say it should be run "within the build folder". If so will this work?:
```
./global/test/TestCommandLineArgumentsTutorial <args>
```
Test it first(!)

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+ // NOTE: These tests will run if executed from terminal directly with their default varaible values. However, these
+ // tutorial tests are intended to be executed with the accompanying runcommandlinetutorial.sh bash script.
+ // Investiagte the accompanying runcommandlinetutorial.sh bash script to see how this tutorial should be executed.
+class TestCommandLineArgumentsTutorial : public CxxTest::TestSuite
+{
+public:
+
+    /* The Default test should be ran with the following bash script starting from line 16.
+    runcommandlinetutorial.sh by default has this script already. However, if you have changed
+    the bash script for another text simply copy and paste between START and END back into the file
```
mirams:
```suggestion
    the bash script for another text simply copy and paste the below code into a shell script (`.sh`) file:
```

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+        for ((k = 2001; k <= M; k += 1000)); do
+          # As bash cannot directly handle double floating point arithmetic we will utilise
+          # awk to divide our varaibles and convert them to doubles.
+           idouble=$(awk "BEGIN {printf \"%.2f\",$i/120}")
+           jdouble=$(awk "BEGIN {printf \"%.2f\",$j/150}")
+           kdouble=$(awk "BEGIN {printf \"%.2f\",$k/180}")
+
+           # Each loop runs an instance of the TestCommandLineArgumentsTutorial with a vector
+           # containing the double corrected version of our varaibles.
+           ~/build/global/test/TestCommandLineArgumentsTutorial --my-vector-of-arguments $idouble $jdouble $kdouble &
```
mirams: same as above

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+        unsigned sum = outp1 + outp2 + outp3;
+        
+        std::cout << "When we add "<< outp1 << " ,"<<  outp2 << " and "<< outp3 <<" we get " << sum <<"\n";
+    }
+    
+    /*
+      In addition to using unsigned integers as command line arguments we can also pass in both doubles and a vector of arguments.
+      This could be very useful for several reasons, such as having a large list of parameters one may wish to pass into a test. 
+
+      For this test change the runcommandlinetutorial.sh bash script from line 16 onwards as follow by copy and pasting between START and END.
```
mirams:
```suggestion
      To script the running of this test with lots of different arguments, copy and paste the following code into a bash script (`.sh`) file.
```

global/test/TestCommandLineArgumentsTutorial.hpp
```diff
@@ -0,0 +1,173 @@
...
+ */
+
+ // NOTE: These tests will run if executed from terminal directly with their default varaible values. However, these
+ // tutorial tests are intended to be executed with the accompanying runcommandlinetutorial.sh bash script.
+ // Investiagte the accompanying runcommandlinetutorial.sh bash script to see how this tutorial should be executed.
+class TestCommandLineArgumentsTutorial : public CxxTest::TestSuite
+{
+public:
+
+    /* The Default test should be ran with the following bash script starting from line 16.
```
mirams: I'd suggest removing bash script and altering this to give tutorial instructions to copy and paste the whole code block below into a new script inside the build folder.

mirams (changes requested):
A few tweaks @BJackal I think it is getting there, the main one is probably to just get the tutorial to put the bash script in a code block, and give instructions to copy and paste into a bash script in the build folder, then you can remove the script and the need for hardcoded paths I think.

mirams:
(you can manually run the generate tutorials and website build to see how the generated webpage looks following these instructions: https://github.com/Chaste/Chaste/issues/193#issuecomment-1908057374)

global/runcommandlinetutorial.sh
BJackal: With the code block you suggested instead carrying out this bash script it makes sense to remove it for tidying purposes I think. Its possibly also a good way to set up best practice for such bash scripts if there isn't already one. So there aren't multiple bash scripts floating around.

global/test/TestCommandLineArgumentsTutorial.hpp
BJackal: It could be removed but I think I could also add some basic testing style structure here ? As other tutorials seem to have some small additions for testing purposes. I suppose it could also act as a way to ensure nothing odd happens. I'm also with for removing it though.

mirams (approved):

mirams:
Didn't mean to delete branch, we'll do that when ticket is closed!

We'll want to add (manually) an entry to the "Core Functionality" part of the menu page: https://chaste.github.io/docs/user-tutorials/

BJackal:
@mirams. Made a branch in Chate.github.io for including the command line arguments tutorial under core functionality and made a pull request.

mirams:
OK, closed by https://github.com/Chaste/Chaste.github.io/pull/95

PR #301, opened 2024-08-28
Close issue 300 Sampling timestep multiple with outputting cell velocity

### LINKED ISSUE #300: FIx output timesteps for Velocity Output in Cell Based simulations
Currently when mOutputCellVelocities is true in AbstractCellBasedSimulation the velocities are output on mSamplingTimestepMultiple+1 when they should be output on mSamplingTimestepMultiple. 

This means that velocities arent output on the last timestep.

jmosborne:
This will close the issue as now the putput step is calculate correctly. 

Note this actually passes all the tests without any changes as the cell velocity was only output at every timestep so I also updated a test to only output evert 12 steps.


jmosborne:
Sorry Fergus added you or this super trivial simple pull request:)

mirams (approved):

fcooper8472 (approved):

PR #303, opened 2024-08-28
Finalise PETSc vector assembly after VecSetValues

### LINKED ISSUE #302: PETSc 3.19 vector assembly
While working on Ubuntu 24.04 support (#257), the following was misidentified as being related to #249 but is a separate PETSc-related problem.

v3.19 introduced a change (https://petsc.org/release/changes/319/):

> Enforce the rule that VecAssemblyBegin() and VecAssemblyEnd() must be called on even sequential vectors after calls to VecSetValues().

This was leading to test failures:

```
The following tests FAILED:
	191 - TestCompressibleNonlinearElasticitySolver (Failed)
	199 - TestStokesFlowSolver (Failed)
```

The Chaste function `PetscVecTools::Finalise` is responsible for applying these two methods:
https://github.com/Chaste/Chaste/blob/60e837ec54cfdf0e6b6cce5c8a3e037d883c9a61/linalg/src/PetscVecTools.cpp#L48-L52

But as things stand there are several instances of unfinalized calls to `VecSetValues` in `PetscVecTools`.

Adding a call to `Finalise` after each of these calls resolves the test failures, but it's possible that this would introduce regressions if there were ever many calls to `PetscVecTools::AddToElement` in a tight loop.

Perhaps @mirams and @jmpf could look at the associated PR and comment if we need a different approach.

fcooper8472:
See #302

### DIFF
--- a/linalg/src/PetscVecTools.cpp
+++ b/linalg/src/PetscVecTools.cpp
@@ -59,6 +59,7 @@ void PetscVecTools::SetElement(Vec vector, PetscInt row, double value)
     if (row >= lo && row < hi)
     {
         VecSetValues(vector, 1, &row, &value, INSERT_VALUES);
+        Finalise(vector);
     }
 }
 
@@ -70,6 +71,7 @@ void PetscVecTools::AddToElement(Vec vector, PetscInt row, double value)
     if (row >= lo && row < hi)
     {
         VecSetValues(vector, 1, &row, &value, ADD_VALUES);
+        Finalise(vector);
     }
 }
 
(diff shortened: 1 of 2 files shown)


jmpf (changes requested):
The point of Finalise is to keep PETSc efficient.  Primarily its use is in sparse matrix assignment where it's important to block allocate all the non-zero elements.  I'm not sure how relevant it is with vectors (which are not sparse).

It would be preferred if these Finalise steps only happen at the moment when the vector mode change is actually needed.

PR #321, opened 2024-10-01
Downgrade libexpat from all 2.4.7-1ubuntuX.Y

### LINKED ISSUE #320: Ubuntu 22.04 libexpat downgrade tweak
Related to #249 

@ptheywood added the following to our relevant CI workflows:

```bash
if dpkg -l | grep libexpat1-dev | grep 2.4.7-1ubuntu0.3; then
  sudo apt-get install -y --allow-downgrades libexpat1=2.4.7-1 libexpat1-dev=2.4.7-1
fi
```

but builds are currently failing again because Ubuntu have patched to version 0.4, so the workflow no longer recognises the need to downgrade.

For now I will simply change the second grep to the more permissive `2.4.7-1ubuntu` which will immediately fix this in the short term.

The longer-term fix is to modify the `chaste-dependencies` package for Ubuntu 22.04 to pull in the patched VTK 9.3, but I will open a separate issue for that.

fcooper8472:


mirams (approved):
no idea, approved!

PR #326, opened 2024-10-08
Update issue templates

mirams:
Add a new template issue that reminds us what needs doing for a new dependency


mirams:
@kwabenantim - you could add some links to advice for setting up a new dependency module, running tests with it, and adding to regular test rotation. Would be handy

kwabenantim (approved):

PR #336, opened 2024-10-22
Add GPU support to chaste

### LINKED ISSUE #265: Enhance GPGPU functionality using FLAME GPU
WP 2.4 of the BBR grant is to "Enhance GPGPU functionality using FLAME GPU":

> We will also significantly enhance Chaste GPGPU functionality by utilising [FLAME GPU](https://github.com/FLAMEGPU/FLAMEGPU2), an open-source agent-based simulation middleware that provides a transparent mapping between agent specifications and optimised CUDA GPU code (Milestone 2.4: Enhance GPGPU functionality using FLAME GPU, months 41-60).
> 
> Based on preliminary work, we anticipate an order-of-magnitude speed up incell-based simulations under these developments.

The planned process for this work is to:

1. Develop a minimal FLAME GPU implementation of the 2D vertex model from Chaste (or th 2D overlapping spheres model)
2. Integrate FLAME GPU with existing Chaste build infrastructure
3. Develop a minimal FLAME GPU accelerated overalpping spheres or vertex model within Chaste
4. Expand features of the FLAME GPU accelerated model, e.g. incorporating additional cell cycle models, forces etc. 

These individual parts will be completed as separate issues/tasks, with this issue providing an overview for the workpackage

### 1. Minimal FLAME GPU 2D vertex model prototype

Initially, a prototype implementation of a Chaste Model will be implemented in FLAME GPU 2. 

This will be created in a separate independent repository, as the code is not intended to make it's way into Chaste/Chaste until a latter step.

This repository will:

+ Be a CMake project which
  + Fetches a specific version of [FLAMEGPU/FLAMEGPU2](https://github.com/FLAMEGPU/FLAMEGPU2) via CMake FetchContent at configuration time
  + Include a FLAME GPU 2 implementation of `Test2DVertexCryptRepresentativeSimulationForProfiling` from `crypt/test/Test2DVertexBasedCryptRepresentativeSimulation`
  + Include some unit tests
  + Potentially include fetching a copy of Chaste/Chaste to re-use host-components, i.e. disk-io (subject to complexity/requirements)

This may then be extended to include some rough benchmarking to compare the performance of this test case at a range of model scales

+ Adapting interface to allow model scale to be selected at runtime
+ Script(s) for benchmarking the FLAME GPU implementation at a number of model scales
+ Scripts for performance data extraction and plotting
+ Possibly a modified copy of `Test2DVertexCryptRepresentativeSimulationForProfiling`
  + Adapted for use in a scaling benchmark, i.e. with a CLI allowing the model scale to be set, and possibly finer-grained performance capture

The source code for this prototype will be structured to mirror the Chaste API and model structure closely, so guide the subsequent integration steps. 

### 2. Integrate FLAME GPU 2 with existing Chaste Build infrastructure

For non-python FLAME GPU 2, there is not currently a mechanism for installing FLAME GPU 2 system wide, so this will likely be handled via CMake using Fetch Content. 
Changes to make FLAME GPU 2 installable (and therefore not needing to be fetched per configure) are planned for a future FLAME GPU 2 release ([FLAMEGPU/FLAMEGPU2#604](https://github.com/FLAMEGPU/FLAMEGPU2/issues/604))

It probably makes sense for this to be in a long-life branch on Chaste/Chaste (or a fork) rather than hitting `develop` for the duration of the work package. I.e. `gpu-develop` or similar?

This will also need to be an opt-in feature for reasons listed below:

+ CMake option(s) to make this an opt-in feature due to software/hardware requirements
    + CUDA >= 11.0 requirement (potentially >= 11.2 in the near future)
    + Compute Capabiltiy >= 35 (2nd generation Kepler) GPU required for CUDA 11.x builds
    + Compute Capability >= 50 (Maxwell) GPU required for CUDA 12.x builds 
    + Reduced OS compatibility
        + FLAME GPU Supports Linux and Windows. 
        + CUDA is not avilable on recent Mac OS due to the lack of Nvidia GPUs in Macs. 
    + Reduced compiler compatibility
        + Currently only GCC is supported by FLAME GPU 2 (Version compatible with appropriate NVCC and C++17, so GCC 8?)
        + Clang support is partial, with a draft PR to improve this ([FLAMEGPU/FLAMEGPU2#623](https://github.com/FLAMEGPU/FLAMEGPU2/issues/623), [FLAMEGPU/FLAMEGPU2#936](https://github.com/FLAMEGPU/FLAMEGPU2/pull/936))
        + NVHPC++ support is partial, with a draft PR to improve this ([FLAMEGPU/FLAMEGPU2#977](https://github.com/FLAMEGPU/FLAMEGPU2/issues/977), [FLAMEGPU/FLAMEGPU2#693](https://github.com/FLAMEGPU/FLAMEGPU2/pull/693))
        + ICC has not been tested as a host compiler for FLAME GPU 2, so may or may not work ([FLAMEGPU/FLAMEGPU2#1202](https://github.com/FLAMEGPU/FLAMEGPU2/issues/1202))
+ Use [FetchContent](https://cmake.org/cmake/help/v3.16/module/FetchContent.html) to clone/download a versioned commit of FLAMEGPU/FLAMEPGU2 and make it avilable for use during CMake (by linking gainst the FLAMEGPU target, and optionally using some other CMake methods)
+ CI:
    + Codecov does not support CUDA, so may wish to prevent this combinition in CI
    + CI workflows
        + Free/default GitHub Actions runners do not include GPUs. NVCC can be installed within the workflow runner (or a container could be used) to check for compilation, but tests would not be executable in these workflows
        + Paid-for GPU runners (Nvidia T4) are now available in a [public beta as of 2024-04-02](https://github.blog/changelog/2024-04-02-whats-new-for-github-actions-hosted-runners/), which would enable running GPU tests in CI (with cost). 
    + Docker Container limitiations
        + Due to `libcuda.so` licence restrictions, any container distributing `libcuda.so` needs to be based on an nvidia provided container. 
        Ubuntu base images are provided, so this shouldn't be too much of an issue. 
        Containers Images will however be quite large.
    + Memory Checking
       + Valgrind (used to) produces many false positives during leak checking for CUDA applications, requiring a customise suppression list, which may be required.
       + [CUDAs memory checking tools](https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html#memcheck-tool) may be beneificaial to use in addition (manually, unless a GPU-enabled CI becomes available). 

### 3. Develop a minimal FLAME GPU accelerated vertex model within Chaste (or overalpping spheres)

Once the 2D Vertex model has been prototyped, and it is possible to optionally include/depend on FLAME GPU 2 within Chaste's CMake, FLAME GPU will need to be integrated within Chaste's API itself.

I.e. Users should just use the regular Chaste API with some form of GPU opt-in for GPU enabled builds (or just for any GPU enabled build) where possible. 

The exact shpae of this will be guided by the Prototye implementation, which intends to mirror the structure of Chaste models closesly. 

Prior to the prototyping it appears that:

+ Chaste Simulation classes (i.e. `CryptSimulation2d`) will roughly need to contain the FLAME GPU `model` and `simulation` classes
+ FLAME GPU 2 `Agent` and `Populations` can only exist for a specific model, so the FLAME GPU 2 equivalents may only be possible after they have been passed to the Chaste simulation
  + Due to the lack of introspection in C+, this might be the challenging part
+ Concentrations (I.e. `WntConcentration`) will likely want implementing via [FLAME GPU 2 Macro Environment Properties](https://docs.flamegpu.com/guide/environment/index.html#defining-macro-properties), however, use of atomic operations may cause non-determinism, leading to an agent-based implementation being preferable (but slower). 
+ Chaste Cell Population classe (i.e. `VertexBasedCellPopulation`) will need to contain/producve a FLAME GPU 2 `Population` ()
+ `simulator.Solve()` will need to do the final construction of the FLAME GPU Simulation object(s) and data setting, prior to calling it's `simulate()` method. 
+ An additional call to the static method `flamegpu::cleanup()` may be benficial, but is not strictly required (at the cost of memory leaks), however, it should only be called once. 

### 4. Expand features of the FLAME GPU accelerated model

Once the minimal 2D vertex model has been integrated into a branch in Chaste, the set of GPU supported modeling features will need to be expanded to a broader set of Chaste's features.

More thorough testing and validation will likely also be desired / worth while (i.e. Floating point values are unlikely to be exactly comparable).

MILeach:
This PR relates to #265 and adds GPU support for elements of chaste simulations.

# Meeting minutes 29/10/2024
## Proof of concept
- Agreed that targeting force resolution & integration is a sensible next step
- Aim is to produce a minimal working example using FLAME GPU to perform force calculations
- On updating FLAME simulation data vs 'clean slate' each time step - better to restart the flame sim each step by clearing agent count and pulling all the data fresh from chaste. Clearing agent count will preserve the allocated memory and avoid allocation costs per timestep. Also leaves open future possibility of running multiple physics steps per biological step without having to use a submodel. The FLAME model can just run until some condition is met.
- Minimal model can be produced by embedding the FLAME simulation in a modifier. The modifier performs all the functionality that the force class and time stepper usually would. No force is added to the chaste sim via the conventional method, meaning the CPU will not perform any force calculations. 
- FLAME model definition can take place during the simulation modifier construction
- Use a test which does some simple force calculations using both chaste and flame to compare results & demonstrate equivalence
- Expand to perform full force calculations in the simulation then benchmark against CPU for different sizes of population

## Long term integration
- Write up options for this in report on GPU acceleration of Chaste
  - How to handle construction of GPU model?
  - How much should the user be involved vs automation?
  - Where should the FLAME model / CUDASimulation objects live?
  - How to handle data transfers?

# To Do
- [x] Create branch for GPU work
- [x] Modify chaste CMake to pull in FLAME GPU
- [x] Modify chaste CMake to build FLAME GPU
- [x] Add flamegpu target as dependency
- [x] Get chaste to build with nvcc  (cuda compiler)
- [x] Modify cmake to recognise .cu/.cuh files in the chaste build process
- [x] Create a new test file for GPU functionality, which includes flamegpu.h
- [x] Add a test which runs a minimal FLAME GPU simulation
- [x] Create a test which performs equivalent force calculations in chaste and with flame
- [x] Create a modifier which performs the force calculations
- [x] Disable the force calculations & integration in the CPU pathway
- [x] Benchmark
- [ ] Update testing infrastructure

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,7 @@ message ("####################################\n")
 message(STATUS "CMake version: ${CMAKE_VERSION}")
 
 # Since CMake 3.0, version can be supplied to the project() command.  We set it directly, for users on older CMake
-project (Chaste VERSION 2024.2)
+project (Chaste VERSION 2024.2 LANGUAGES CUDA CXX)
 set (Chaste_VERSION_MAJOR 2024 CACHE STRING "Chaste major version number")
 set (Chaste_VERSION_MINOR 2 CACHE STRING "Chaste minor version number")
 
@@ -579,6 +579,32 @@ if (Chaste_ENABLE_TESTING)
     list (APPEND CMAKE_INCLUDE_PATH "${Chaste_SOURCE_DIR}/cxxtest")
     find_package (CxxTest)
 endif ()
+
+####################
+# FIND FLAMEGPU 2 #
+####################
+#t # Minimum CMake version 3.18 for CUDA --std=c++17 
+cmake_minimum_required(VERSION 3.18...3.25 FATAL_ERROR)
+#
+## Optionaly set the version of flamegpu which should be used, ideally a tag (i.e. `v2.0.0-rc`) or branch name, or potentially a commit hash.
+set(FLAMEGPU_VERSION "master" CACHE STRING "FLAMEGPU/FLAMEGPU2 git branch or tag to use")
+## If the above version is a hash instead, also set FLAMEGPU_VERSION_ALLOW_HASH to ON
+# set(FLAMEGPU_VERSION_ALLOW_HASH "ON")
+#
+## Manually specify the FLAMEGPU_VISUALISATION option to provide it prior to original configuration and allow the default to be overridden in the downstream project
+option(FLAMEGPU_VISUALISATION "Enable FLAMEGPU visualisation support" OFF)
+#
+## Our core dependency is FLAMEGPU2 l:warningsib, first lets find it
+include(${CMAKE_CURRENT_LIST_DIR}/cmake/flamegpu2.cmake)
+#
+## Handle CMAKE_CUDA_ARCHITECTURES gracefully, passign the project name for code-injection
+include(${FLAMEGPU_ROOT}/cmake/CUDAArchitectures.cmake)
+flamegpu_init_cuda_architectures(PROJECT Chaste)
+#
+## Include common rules from the FLAMEGPU/FLAMEGPU2 repositories CMake
+include(${FLAMEGPU_ROOT}/cmake/common.cmake)
+#        
+target_link_libraries(Chaste_COMMON_DEPS INTERFACE flamegpu)
 
 ###########################################
 # SETUP AVAILABLE COMPONENTS AND PROJECTS #
--- a/cell_based/src/simulation/modifiers/GPUModifier.cu
+++ b/cell_based/src/simulation/modifiers/GPUModifier.cu
@@ -36,7 +36,7 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "GPUModifier.cuh"
 #include "MeshBasedCellPopulation.hpp"
 
-FLAMEGPU_AGENT_FUNCTION(output_location, flamegpu::MessageNone, flamegpu::MessageBruteForce) {
+FLAMEGPU_AGENT_FUNCTION(output_location, flamegpu::MessageNone, flamegpu::MessageSpatial2D) {
     FLAMEGPU->message_out.setVariable<float>("x", FLAMEGPU->getVariable<float>("x"));
     FLAMEGPU->message_out.setVariable<float>("y", FLAMEGPU->getVariable<float>("y"));
     FLAMEGPU->message_out.setVariable<float>("radius", FLAMEGPU->getVariable<float>("radius"));
@@ -44,14 +44,14 @@ FLAMEGPU_AGENT_FUNCTION(output_location, flamegpu::MessageNone, flamegpu::Messag
 }
 
 // Models repulsion force without division/apoptosis
-FLAMEGPU_AGENT_FUNCTION(compute_force_meineke_spring, flamegpu::MessageBruteForce, flamegpu::MessageNone) {
-    const float x = FLAMEGPU->getVariable<float>("x");
-    const float y = FLAMEGPU->getVariable<float>("y");
+FLAMEGPU_AGENT_FUNCTION(compute_force_meineke_spring, flamegpu::MessageSpatial2D, flamegpu::MessageNone) {
+    const double x = FLAMEGPU->getVariable<float>("x");
+    const double y = FLAMEGPU->getVariable<float>("y");
     float x_force = 0.0;
     float y_force = 0.0;
     float radius = FLAMEGPU->getVariable<float>("radius");
 
-    for (const auto& message : FLAMEGPU->message_in) {
+    for (const auto& message : FLAMEGPU->message_in(x, y)) {
         float other_x = message.getVariable<float>("x");
         float other_y = message.getVariable<float>("y");
         float other_radius = message.getVariable<float>("radius");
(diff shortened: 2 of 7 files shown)


MILeach:
## GPU Modifier progress 12/09/2024

A modifier has been added in `src/simulation/modifiers/GPUModifier.{cu|cuh}` which offloads force computation to the GPU. This presently supports a single force model, in 2 dimensions and performs an euler integration step at the end of the force computation. The `SetupSolve` method defines the FLAME GPU model and the `UpdateAtEndOfTimestep` method handles transfer of data and running the FLAME GPU simulation. The simulation is reset, and all agent data is transferred in full in each direction in each step, i.e. there is no partial update of only 'dirty' data.

### Tasks for modifier implementation
- [x] Create a new `GPUModifier` class
- [x] Initialise FLAME GPU model inside `SetupSolve`
- [x] Define FLAME GPU agent functions which model force computation
- [x] Wrangle chaste node data to generate an equivalent FLAME GPU population
- [x] Run the simulation
- [x] Wrangle the data back to chaste
- [x] Create a test which uses the modifier
- [ ] Ensure CPU pathway for force calculation & integration is definitely not being used
- [x] Compare the test results & benchmark against an equivalent CPU-based chaste simulation
- [x] Create an alternate agent function which uses spatial messaging instead of brute force messaging & re-benchmark
- [ ] Add support for a boundary condition type
- [ ] Add support for 1 & 3 dimensions
- [ ] Refactor agent functions into a library of composable pieces
- [ ] (Longer term) Expand support for other types of force models etc.

MILeach:
## Preliminary Results
- Overall speedup ~4x
- All GPU activities (collection & transfer of data + simulation) take ~ 1/8th of the total run time per time step
- Suggests ~ 24x speedup of force calculations
- Assuming flame cpu calls are blocking until related gpu operations are complete, rough breakdown of gpu time:  
  - Wrangling data to GPU: 60%
  - Simulation: 10%
  - Wrangling data back to CPU: 30%
  - Can complete ~300 physics time steps per 'biological' timestep with equal performance

PR #347, opened 2025-01-06
Move to chaste/base image for devcontainer

fcooper8472:
When GitHub creates a new codespace, there is an intermediate (undocumented) provisioning step, with less space available than the resulting codespace machine.

As a result, using the larger pre-built image doesn't allow users to create a default Chaste codespace.

This PR changes the default to chaste/base which allows creation of codespaces.


kwabenantim (approved):
Looks good to me!

bdevans (approved):
Looks good to me too. We should make a note to change it back after the workshop however so that it's setup for the usual dev workflow.

PR #350, opened 2025-01-06
Use root user in devcontainer

fcooper8472:
Temporary solution to a permissions issue in the new base image


fcooper8472:
Closing as no longer relevant.

Issue #353, opened 2025-01-09
TestCardiacSimulationX tests fails when run in parallel


MichaelClerx:
**Describe the bug**
Running the tests below fails with -jN and N>1, with the number of failures increasing as N gets larger.

**To Reproduce**

Load chaste docker with
```
docker run --init -it --rm -v chaste_data:/home/chaste chaste/release
```
then
```
chaste@a5db1df53490:~/build$ ctest -R TestCardiacSimulation -E TestCardiacSimulationN -j8
Test project /home/chaste/build
    Start 445: TestCardiacSimulation
    Start 533: TestCardiacSimulationCodegen
    Start 534: TestCardiacSimulationArchiverCodegen
    Start 446: TestCardiacSimulationArchiver
    Start 502: TestCardiacSimulationArchiverParallel
1/5 Test #534: TestCardiacSimulationArchiverCodegen ....***Failed    1.85 sec
2/5 Test #446: TestCardiacSimulationArchiver ...........***Failed    1.86 sec
3/5 Test #445: TestCardiacSimulation ...................Subprocess aborted***Exception:   4.87 sec
4/5 Test #502: TestCardiacSimulationArchiverParallel ...   Passed    6.69 sec
5/5 Test #533: TestCardiacSimulationCodegen ............***Failed   21.06 sec

20% tests passed, 4 tests failed out of 5

Label Time Summary:
Codegen_heart       =  22.91 sec*proc (2 tests)
Continuous_heart    =   6.72 sec*proc (2 tests)
Parallel_heart      =  13.37 sec*proc (1 test)

Total Test time (real) =  21.07 sec

The following tests FAILED:
	445 - TestCardiacSimulation (Subprocess aborted)
	446 - TestCardiacSimulationArchiver (Failed)
	533 - TestCardiacSimulationCodegen (Failed)
	534 - TestCardiacSimulationArchiverCodegen (Failed)
Errors while running CTest
Output from these tests are in: /home/chaste/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```

but

```
chaste@a5db1df53490:~/build$ ctest -R TestCardiacSimulation -E TestCardiacSimulationN -j1
Test project /home/chaste/build
    Start 445: TestCardiacSimulation
1/5 Test #445: TestCardiacSimulation ...................   Passed   21.00 sec
    Start 446: TestCardiacSimulationArchiver
[...]

PR #355, opened 2025-01-10
Fix duplicate Codegen tests

kwabenantim:
Supports #353


mirams (approved):

kwabenantim:
It seems the cmake for `Codegen` that was mirroring the `Parallel` section was a workaround for adding tests to multiple test packs by renaming the tests. Taking it out made `ctest -L Codegen` empty because all `Codegen` tests were already in `heart`. This has now been fixed by allowing tests to have multiple labels so they can belong to multiple test packs at the same time.

PR #365, opened 2025-02-03
Add a ref to choose specific project commit

### LINKED ISSUE #364: Add a ref for project tests to control which commit is tested
Currently the [project tests](https://github.com/Chaste/Chaste/actions/workflows/project-tests.yml) are failing because the [CellBasedComparison2017](https://github.com/Chaste/CellBasedComparison2017) was updated with respect to PR #313 which hasn't been merged yet.

We should support choosing a specific ref on the project that is being tested, to choose which specific commit is expected to work with the current version of Chaste.

Normally, this could be left as an empty string, which then checks out the default branch. This is currently the behaviour. But, when required, a specific commit hash (or tag, branch etc) could be specified instead.

- [ ] Add a `ref` parameter for user projects in the `project-tests` workflow

fcooper8472:
Closes #364


kwabenantim (approved):
Looks good to me!

PR #373, opened 2025-02-05
372 beef up archiving tests to use separate scopes for save and load where possible

### LINKED ISSUE #372: Beef up archiving tests to use separate scopes for save and load where possible
As per https://chaste.github.io/docs/user-guides/boost-serialization/ it is nicest if saving and loading is tested like this:

```cpp
void TestArchiving()
{
    {
        // Make everything and save the archive

        // We should generally archive the most-abstract pointer of any concrete class
    }
    {
        // Load the abstract pointer archive and test it contains the right thing
    }
}
```
where the only things to exist at the top level scope are numbers to test against etc.

I've found examples where this isn't the case, so this ticket should be a top level check that we are doing that strategy wherever possible.

mirams:
Should close #372

### DIFF
--- a/cell_based/test/cell/TestCellPropertyCollection.hpp
+++ b/cell_based/test/cell/TestCellPropertyCollection.hpp
@@ -145,50 +145,53 @@ public:
         // Archive a cell property collection
         {
             // Create a cell property collection
-            CellPropertyCollection collection;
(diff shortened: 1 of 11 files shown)


fcooper8472 (approved):
Looks good to me!

mirams:
OK, since it is only failing PyChaste and project I'll merge it in.

PR #385, opened 2025-03-04
Improve ESF calculation

### LINKED ISSUE #384: Elongation shape factor might be emitting NAN values
Related to: https://github.com/Chaste/Chaste/discussions/361#discussioncomment-12386869

The Elongation Shape Factor (ESF) is defined as the square root of the ratio of the largest and smallest eigenvalues of the moment matrix. This matrix is positive semi-definite, and the eigenvalues are (mathematically) guaranteed to be non-negative.

That said, it's not impossible that due to floating point arithmetic, a shape could exist that would result in the smallest eigenvalue being very small and negative, which would cause the ESF to be NAN.

It is also possible that, if a shape is (very nearly) colinear, the ESF could be (floating point) infinite. This would also cause problems from an archiving perspective.

Suggestions:

- [x] Ensure it's impossible for the ESF to be NAN
- [x] Warn if the ESF is infinite (which is valid, but might still be annoying)

fcooper8472:
See #384

### DIFF
--- a/mesh/test/vertex/TestVertexMesh.hpp
+++ b/mesh/test/vertex/TestVertexMesh.hpp
@@ -1717,14 +1717,14 @@ public:
 
     void TestInfiniteElongationShapeFactor()
     {
-        // Test method with a single colinear element
-        std::vector<Node<2>*> colinear_nodes;
-        colinear_nodes.push_back(new Node<2>(0, false, 1.0, 0.0));
-        colinear_nodes.push_back(new Node<2>(1, false, 2.0, 0.0));
-        colinear_nodes.push_back(new Node<2>(2, false, 3.0, -1e-12));
-        std::vector<VertexElement<2,2>*> colinear_element;
-        colinear_element.push_back(new VertexElement<2,2>(0, colinear_nodes));
-        VertexMesh<2,2> mesh(colinear_nodes, colinear_element);
+        // Test method with a single collinear element
+        std::vector<Node<2>*> collinear_nodes;
+        collinear_nodes.push_back(new Node<2>(0, false, 1.0, 0.0));
+        collinear_nodes.push_back(new Node<2>(1, false, 2.0, 0.0));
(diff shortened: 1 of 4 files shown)


fcooper8472:
@jmpf when you look at this you might want to hit the cog -> hide whitespace, due to the `if constexpr` indenting the whole block.

mesh/test/vertex/TestVertexMesh.hpp
```diff
@@ -1715,6 +1715,22 @@ class TestVertexMesh : public CxxTest::TestSuite
...
     }
 
+    void TestInfiniteElongationShapeFactor()
+    {
+        // Test method with a single colinear element
+        std::vector<Node<2>*> colinear_nodes;
+        colinear_nodes.push_back(new Node<2>(0, false, 1.0, 0.0));
+        colinear_nodes.push_back(new Node<2>(1, false, 2.0, 0.0));
+        colinear_nodes.push_back(new Node<2>(2, false, 3.0, -1e-12));
+        std::vector<VertexElement<2,2>*> colinear_element;
```
jmpf: "collinear" - it really does have a double l.

jmpf (approved):
Looks good. Did failing test get you a NaN before?

mesh/test/vertex/TestVertexMesh.hpp
Reshief: In my simulations it occasionally returned a NaN, yes. The previous tests did, however, not trigger that case.

fcooper8472:
> Looks good. Did failing test get you a NaN before?

I couldn't find a shape that gave NaN values, no. @Reshief do you have an example of coordinates that generate a NaN value for testing this?

Reshief:
>@Reshief do you have an example of coordinates that generate a NaN value for testing this?

@fcooper8472 
I'm currently building a minimal sample of the system breaking at https://github.com/Reshief/Chaste_MinimalBreakingSample 
It could also be an issue of the random mesh generator as far as I can tell.

PR #386, opened 2025-03-04
Adding functionality to allow for writing of binary archives with CellBasedSimulationArchiver class

### LINKED ISSUE #381: Consider allowing (flag for?) binary archives in `CellBasedSimulationArchiver`
For smaller file sizes and faster loading in production run simulations...

Reshief:
Hi all, 

I have never done a pull request before, so please excuse me, if I missed something in your guidelines on first read.
Because I had issues with serialization in text based archives myself, I had an interest in this being easily accessible in Chase, so I gave it a try:
This PR relates to https://github.com/Chaste/Chaste/issues/381 and proposes an option to implement the ability to write out binary archives in CellBasedSimulationArchiver Save()/Load() functions through a boolean template flag switching from the classical text based archives to binary archives.

More precisely: 
- Adds a template flag to CellBasedSimulationArchiver to allow for binary archive input/output, while keeping prior behavior and keeping code changes to a minimum
- Adds a specialization for ArchiveOpener for binary input/output archive to facilitate minimum code changes in CellBasedSimulationArchiver class
- Adds a test in TestArchivingHelperClasses.hpp to check for correct function of the ArchiveOpener specializations.

As a consequence and as mentioned in the issue #381, this would lead to
- Smaller file sizes of archives
- Allow users to easily circumvent (de)serialization issues associated with NaN values in double fields
- Closes #381 

I assume there may be other options to deal with the binary reading/writing as for example an optional argument to Save() or Load in the CellBasedSimulationArchiver  instead of a template argument. I am open to suggestions, which option you would prefer and I can implement that alternative behavior if it would help you.


cell_based/src/simulation/CellBasedSimulationArchiver.hpp
```diff
@@ -54,8 +54,13 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
  * of all the various AbstractCellBasedSimulation objects. It has no explicit constructor
  * (just uses a default one) and no member variables.
+ * 
+ * The template boolean flag ARCHIVE_IS_TEXT is allows for reading and writing of binary 
+ * archives instead of text-based archives by setting ARCHIVE_IS_TEXT to false.
+ * Default behavior is to output text-based archives for easier debugging and 
+ * human readability.
  */
-template<unsigned ELEMENT_DIM, class SIM, unsigned SPACE_DIM=ELEMENT_DIM>
+template<unsigned ELEMENT_DIM, class SIM, bool ARCHIVE_IS_TEXT=true, unsigned SPACE_DIM=ELEMENT_DIM>
```
mirams: We should probably make the new argument the last one, in case existing code is using the first two. If we do want to do this with a template. My instinct would have been just a flag in the constructor of the CellBasedSimulationArchiver maybe, but I will let @fcooper8472 and @jmpf give input on that.

Reshief: That would have been ideal, however, the Save() and Load() functions are static functions on the CellBasedSimulationArchiver class. So people will not call the constructor to invoke them. 
So for me, the choice was between a template argument or optional argument to the function call itself. 

But I agree, that the new template argument should probably be last to not break compatibility if it stays a template argument.

global/src/checkpointing/ArchiveOpener.cpp
```diff
@@ -209,3 +209,168 @@ ArchiveOpener<boost::archive::text_oarchive, std::ofstream>::~ArchiveOpener()
      */
     PetscTools::Barrier("~ArchiveOpener");
 }
+
```
mirams, line 146: This feels like a lot of replicated code, any way to template the ArchiveOpener over the archive type instead?

cell_based/src/simulation/CellBasedSimulationArchiver.hpp
fcooper8472: @Reshief apologies for the delay, but I've looked into this now and I agree that adding a template argument is probably the right way to go.

Agreed, also, that putting the new template parameter last is the way to go.

Could the template argument be the archive type itself, defaulting to `boost::archive::text_iarchive` (or `boost::archive::text_oarchive` for `::Save`, then meaning we don't need the `if constexpr` block at all?

global/src/checkpointing/ArchiveOpener.cpp
fcooper8472: @Reshief similar to above, this change looks like it's identical to the existing specialization, but with `boost::archive::binary_iarchive` instead of `boost::archive::text_iarchive`.

Could this method also be templated over the archive type, again defaulting to `boost::archive::text_iarchive`?

I've not tested whether that would work, but it would be ideal if we could avoid having to maintain essentially the same code twice, as there's a real disk of divergence in the future.

cell_based/src/simulation/CellBasedSimulationArchiver.hpp
Reshief: I have reworked the SimulationArchiver to have the archive types as template arguments and added two specialization types to help with setting both input and output archive types to either text or binary together. 
This should reduce code duplication.

global/src/checkpointing/ArchiveOpener.cpp
Reshief: I can certainly try. I think, here, the issue is, that the existing implementation for the archive is exclusively provided for the text based archives and not for a generic archive. Let me see, if I can make it work with a generic archive template parameter

Reshief:
I have revised the pull request to reduce code duplication, while still maintaining the addition of the option to have both binary and text formats either through template parameters or type name aliases to make it easier to switch between text and binary

fcooper8472 (commented):
@Reshief with apologies for the delay, I've managed to have a thorough look through this now and I really like this approach.

I would like to see whether it passes all the workflows, but I'm not sure how to kick them all off.

I'm leaving this as a comment rather than approval for now, pending running the workflows.

But thanks for this work - it looks great.

fcooper8472 (dismissed):

fcooper8472:
@Reshief very strange, we can't work out what's preventing us from approving the CI to run our testing workflows.

The only thing I can suggest is can you please try updating your branch with respect to our latest commits on develop? It may be that a merge conflict is preventing the workflows from being runnable.

fcooper8472:
@Reshief just pinging again in case you can update this PR wrt to latest develop branch? I would like to merge this once we can get the workflows to run.

Reshief:
> @Reshief just pinging again in case you can update this PR wrt to latest develop branch? I would like to merge this once we can get the workflows to run.

Sorry, I got a bit sidetracked with work. 
I have merged the Chaste/Chaste develop branch into my branch. 
I hope it helps with the workflow.
Or do I need to do more to make the workflow work on your end?
I didn't touch any of the configuration for CI, so I don't see why it should break.

mirams:
I think it may just be that this PR is too old for the automatic test running. We could try making a new one maybe

kwabenantim:
> I think it may just be that this PR is too old for the automatic test running. We could try making a new one maybe

It looks like the tests wanted approving to start running. It's strange that everything fails though. Possibly, it needs merging the current develop back in.

Reshief:
okay, I'll try and fix the build issue.

Reshief:
Okay, I had some issues with partial class member specialization and now I am running into an issue with the private stream of the stream opener having been freed outside of the destructor of ArchiveOpener. I will try and figure that out.

global/src/checkpointing/ArchiveOpener.cpp
Reshief: So I stumbled upon an issue, namely that Cpp does not allow partial spezialization of member functions.
I have to create a partially specialized subclass and them implement the respective methods. 
Only then can I instantiate them. 
This cuts down on the duplicated logic, but doubles the declaration of shared code like the members of the ArchiveOpener class. 

I assume that duplication of members is preferrable to duplication of logic?

Reshief:
The memory issue was actually a template issue. 
Now the build works and as far as I can tell the tests are passing. Still hasn't finished the last, longer tests, so I will wait for them to finish, but I would say, you can take a look at the changes already. 
It was mostly a fight trying to minimize code duplication while working around the constraints of the C++ template system.

Reshief:
I am not entirely sure what the two failed checks mean. As far as I can read from the erro log, there as an issue related to finding bash on the test runner?
Because all tests I could run locally were passing.

fcooper8472 (approved):
Thanks @Reshief, this looks good to me!

Would be helpful if either @mirams or @jmpf could also take a quick look, this isn't an area of the codebase that I'm especially knowledgeable about.

fcooper8472:
> I am not entirely sure what the two failed checks mean.

@Reshief the two errors I believe are because this is a PR opened from a fork: those runners require access to a secret for uploading a summary of the results to our website. The test suites themselves ran successfully.

mirams (approved):

PR #395, opened 2025-03-21
Remove magic number when setting up box collection

### LINKED ISSUE #394: Remove magic number in NodesOnlyMesh::SetUpBoxCollection
When setting up the box collection in NodesOnlyMesh:

```cpp
    // Note that in circumstances where parts of the bounding box are a long way from the origin then
    // a hard-coded "swell factor" of 1e-14 will not be noticed.
    // For example 200*machine_epsilon = 200*2.2e-16 = 4.4e-14 which implies 200+1e-14 == 200.

    double swell_factor = bounding_box.GetWidth(bounding_box.GetLongestAxis()); // Largest dimension
    swell_factor = (1 + swell_factor) * 1e-14; // Make sure that it's non-zero

    c_vector<double, 2*SPACE_DIM> domain_size;
    for (unsigned i=0; i < SPACE_DIM; i++)
    {
        domain_size[2*i] = bounding_box.rGetLowerCorner()[i] - swell_factor;
        domain_size[2*i+1] = bounding_box.rGetUpperCorner()[i] + swell_factor;
    }
```

I assume this is designed to give a guarantee that every node in the mesh is definitely within the domain of the box collection?

Given the comment that this probably doesn't work even with a fairly realistically-sized mesh, could this be re-worked to remove the magic number by expanding the domain size by "just enough"?

I think `std::nextafter` does exactly what we want.

fcooper8472:
@jmpf see #394 : would like you to sanity-check to make sure I understood what that magic number was previously doing.

### DIFF
--- a/mesh/src/common/NodesOnlyMesh.cpp
+++ b/mesh/src/common/NodesOnlyMesh.cpp
@@ -582,25 +582,19 @@ void NodesOnlyMesh<SPACE_DIM>::SetInitialBoxCollection(const c_vector<double, 2*
     this->SetUpBoxCollection(maxInteractionDistance, domainSize);
 }
 
-template<unsigned SPACE_DIM>
-void NodesOnlyMesh<SPACE_DIM>::SetUpBoxCollection(const std::vector<Node<SPACE_DIM>* >& rNodes)
+template <unsigned SPACE_DIM>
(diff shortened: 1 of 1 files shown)


jmpf (approved):
I believe that, in working on Nick's bug (#95), I added sufficient failing tests to prove that the new more robust code gives the correct behaviour

PR #401, opened 2025-04-01
400 Refactor Cell Populations wrt AbstractTwoBodyInteractionForce

### LINKED ISSUE #400: Refactor cell populations wrt AbstractTwoBodyInteractionForce
At the moment, `AbstractTwoBodyInteractionForce` works with either mesh or node-based cell populations.

The behaviour is different for each.

This is mediated by casting the cell population to check its type and throwing an error if it's not a compatible population. This behaviour isn't quite right. Fundamentally, this abstract force class should work any time the forces are calculable entirely via pairs of nodes. That is a looser constraint than we currently handle.

For instance, the SEM cell population behaves very similarly to the node-based as far as forces are concerned. But adding a third conditional in `AbstractTwoBodyInteractionForce::AddForceContribution` makes this method overly complex.

Given that multiple cell populations all keep track of pairs of nodes, I suggest adding a member:

```cpp
std::vector<std::pair<Node<SPACE_DIM>*, Node<SPACE_DIM>*> > mNodePairs;
```

to the `AbstractOffLatticeCellPopulation`, which will then have a member function `rGetNodePairs()`. This would be virtual in the abstract class and just contain an exception warning that subclasses must override the method.

After this change, any cell population that implements `rGetNodePairs()` will be able to use `AbstractTwoBodyInteractionForce`s.

fcooper8472:
Addresses #400 by modifying the cell population hierarchy so that every off lattice population has an `mNodePairs` member, which cell populations can choose to keep updated if they wish.

Right now this is NodeBased and MeshBased, but will also include SemBased (and potentially ManyVertexBased in future).

Main changes:

- `mNodePairs` moved from `AbstractCentreBasedCellPopulation` up one to `AbstractOffLatticeCellPopulation`
- `rGetNodePairs` now returns that, and it's up to each derived population to decide if it wants to keep that updated or not
- `AbstractTwoBodyInteractionForce` now just iterates over all node pairs, but will emit a warning (only once) if there are no node pairs, suggesting you check whether your cell population is updating node pairs.
- `MeshBasedCellPopulation` now updates node pairs using the spring iterator. This could be rolled out to a few places, e.g. updating rest lengths, because it's a little bit nontrivial keeping track of which springs are visited in the iterator. Calculating only once per timestep will be marginally more efficient than repeatedly using the iterator.

### DIFF
--- a/cell_based/src/population/AbstractCentreBasedCellPopulation.hpp
+++ b/cell_based/src/population/AbstractCentreBasedCellPopulation.hpp
@@ -284,16 +284,6 @@ public:
     virtual bool IsParticle(unsigned index);
 
     /**
-     * Method to return the connected nodes in  a centre based simulation.
-     *
-     * As this method is pure virtual, it must be overridden
-     * in subclasses.
-     *
-     * @return Node pairs for force calculation.
-     */
-    virtual std::vector< std::pair<Node<SPACE_DIM>*, Node<SPACE_DIM>* > >& rGetNodePairs()=0;
-
-    /**
      * @return mMeinekeDivisionSeparation
      */
     double GetMeinekeDivisionSeparation();
--- a/cell_based/src/population/AbstractOffLatticeCellPopulation.cpp
+++ b/cell_based/src/population/AbstractOffLatticeCellPopulation.cpp
@@ -100,6 +100,18 @@ double AbstractOffLatticeCellPopulation<ELEMENT_DIM, SPACE_DIM>::GetAbsoluteMove
 }
 
 template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
+const std::vector<std::pair<Node<SPACE_DIM>*, Node<SPACE_DIM>*> >& AbstractOffLatticeCellPopulation<ELEMENT_DIM, SPACE_DIM>::rGetNodePairs() const
+{
+    EXCEPTION("This method must be overridden in any derived cell population using this functionality.");
+}
+
+template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
+std::vector<std::pair<Node<SPACE_DIM>*, Node<SPACE_DIM>*> >& AbstractOffLatticeCellPopulation<ELEMENT_DIM, SPACE_DIM>::rGetModifiableNodePairs()
+{
+    return mNodePairs;
+}
+
+template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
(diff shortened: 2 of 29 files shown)


fcooper8472:
@jmpf and @mirams I'm expecting some tests to fail, all related to saving and loading mesh based simulations. Not sure why.

But the main structure is ready to review while I iron out what's happening with the failing tests.

jmpf (commented):
I don't like the failing tests.  It's not as simple as it falling over because an archive format has changed (and it's trying to read the old format).  TestOffLatticeSimulation3d writes an archive in one test and then crashes when it attempts to read the same file back.

fcooper8472:
@jmpf I might need some help with this one. I've been completely unable to get to the bottom of the archiving problems.

fcooper8472:
@jmpf and I tracked down what was happening, which was quite subtle.

Initially, the plan was to archive the vector `mNodePairs` in `AbstractOffLatticeCellPopulation`, since that meant not having to recalculate them in the `MeshBasedCellPopulation` constructor when the archive was loaded.

This failed for the reason that there is custom serialization logic for nodes. This could be related to [trac ticket 1709](https://github.com/Chaste/trac_archive/blob/master/issues/1709.md), where older boost could not serialize `c_vector`s.

Nodes, therefore, aren't directly archived, but data necessary to recreate them is. When I added `mNodePairs` to the archive, this worked just fine with modern boost, but there was no direct way of knowing that the de-serialized nodes had nothing to do with the actual nodes serialized with the mesh.

Normally this might have gone completely unnoticed because `mNodePairs` is recalculated during remeshing in the `MeshBasedCellPopulation`. But the specific test that was failing turns off remeshing because it's a 2d-in-3d test:

```cpp
        // Stops remeshing as not possible for 2d in 3d meshes
        simulator.SetUpdateCellPopulationRule(false);
```

So: the force was being correctly calculated and being added to the nodes, but the nodes had nothing to do with the mesh, so were never getting their location updated.

The resolution is to simply not archive the node pairs, as in all cases they can be trivially re-calculated, and in most populations must be recalculated every timestep in any case.

In the case of `MeshBasedCellPopulation` and subclasses, the appropriate place to do this is during serialization, where there is now a call to

```cpp
this->UpdateNodePairs();
```

Technically this also means there is unnecessary work when saving the archive, but the alternative of splitting the archiving into a separate save and load isn't worth the savings.

jmpf (approved):
Looks complete now.  CI failures are spurious.

PR #416, opened 2025-05-28
Fix coverage by fixing parallel behaviour with HDF5 1.12.0

jmpf:


mirams (approved):
You'll have done on purpose I am 90% sure @jmpf , but a slight difference in logic with all the setup now called when `mNumberOwned==0` whereas it was skipped before - is that intended?

jmpf:
> You'll have done on purpose I am 90% sure @jmpf , but a slight difference in logic with all the setup now called when `mNumberOwned==0` whereas it was skipped before - is that intended?

Both changes are the same.

It used to be  
```if (you own data){  normal set up} else /*no owned data*/{ null setup}``` 
but now it's 
```(everybody) { normal set up}; if (no owned data){state that there is no data}```

PR #420, opened 2025-06-18
224 improvements to tutorials

### LINKED ISSUE #224: Improvements to tutorials

mirams:
Just removing a few more mentions of `/tmp/$USER/testoutput` from the source code.

### DIFF
--- a/cell_based/src/simulation/AbstractCellBasedSimulation.hpp
+++ b/cell_based/src/simulation/AbstractCellBasedSimulation.hpp
@@ -131,7 +131,7 @@ protected:
     /** Whether to update the topology of the cell population at each time step (defaults to true).*/
     bool mUpdateCellPopulation;
 
-    /** Output directory (a subfolder of tmp/[USERNAME]/testoutput). */
+    /** Output directory (a subfolder of $CHASTE_TEST_OUTPUT). */
     std::string mOutputDirectory;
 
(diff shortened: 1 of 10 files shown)


kwabenantim (approved):
All these changes look good to me.

I've removed one more from `TestDistributedQuadraticMesh.hpp`.

There's also some example MATLAB code in there as well with a hardcoded path, but that probably needs replacing with the correct path if someone wants to use it, unless we want to recommend something like `getenv("CHASTE_TEST_OUTPUT")` there, which will only work if the MATLAB instance is sharing the same environment with Chaste.

https://github.com/Chaste/Chaste/blob/69fbcc9adf4a02b7ac802c1766aea9a5cca86b2a/mesh/test/TestDistributedQuadraticMesh.hpp#L187

PR #422, opened 2025-06-23
410 migrate to petscs packageconfig based cmake integration and improve mpi discovery

### LINKED ISSUE #410: Migrate to PETSc's PackageConfig-based CMake integration and improve MPI discovery
Our current method of finding PETSc relies on a custom `FindPETSc.cmake` module, which was originally necessary due to historical inconsistencies in PETSc's installation layout and lack of standardised CMake support. This approach has led to a build setup that is now unnecessarily complex and fragile.

PETSc now provides an official [PackageConfig](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html)-based integration, documented here:

* [https://petsc.org/release/faq/#can-i-use-cmake-to-build-my-own-project-that-depends-on-petsc](https://petsc.org/release/faq/#can-i-use-cmake-to-build-my-own-project-that-depends-on-petsc)
* [https://gitlab.com/petsc/petsc/-/blob/main/share/petsc/CMakeLists.txt](https://gitlab.com/petsc/petsc/-/blob/main/share/petsc/CMakeLists.txt)

We should adopt this in place of our current [FindPETSc.cmake](https://github.com/Chaste/Chaste/blob/develop/cmake/Modules/FindPETSc.cmake).

However, our Dependency Modules system still compiles and bundles HDF5 and MPICH alongside PETSc. Therefore, it is critical that our new approach continues to enforce the correct MPI implementation during CMake configuration, to avoid inadvertently using the system MPI when a bundled one is available.

fcooper8472:
- [x] Modernise FindPETSc.cmake
- [x] Bump minimum CMake version from 3.16.3 (Ubuntu 20.04) to 3.22.1 (Ubuntu 22.04)
- [x] Modernise finding dependencies:
  - [x] PETSc
  - [x] HDF5
  - [x] MPI


fcooper8472:
Currently this is failing due to the CMake minimum requirement. This requires a version newer than the default available on Ubuntu 20.04.

See #425 for discussion about turning off Ubuntu 20.04 support.

kwabenantim (approved):
This looks good to me. It finds the correct version of PETSc in all portabilities and also finds the MPI executable path correctly without any add-on PETSc flags.

Issue #424, opened 2025-06-25
Review support for ParMETIS


jmpf:
See #421 where libparmetis-dev is being dropped from Ubuntu 25.04 over licensing issues.

* It may be possible to use libscotchparmetis-dev as a drop in replacement (see #421). Note that licence is also non-standard.
* We should consider the code for `PETSC_MAT_PARTITION` which is probably never used since it requires a PETSc install which is compiled to include ParMETIS
* It may be the time to drop ParMETIS style partitioning
* We may also consider just skipping support for 25.04: https://launchpad.net/ubuntu/+source/parmetis looks as though ParMETIS is proposed for including in Questing Quokka, which presumably is 25.10

jmpf:
Drop in support may not be there.  I think we are using a helper method in parmetis ` ParMETIS_V3_Mesh2Dual` which is not implemented in scotch.  (...Or at least it wasn't in 2011.  What happened since?)
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=653647

jmpf:
Looks like we deprecated the use of METIS and/or ParMETIS nearly a decade ago:
https://github.com/Chaste/Chaste/commit/31f5c57de873f7a39f6c684997c6ae208c736cad
shows that calls to METIS library were deprecated.  However, it seems that (by accident or design) parMETIS was marked as the deprecated one:
https://github.com/Chaste/Chaste/blame/develop/mesh/src/common/DistributedTetrahedralMeshPartitionType.hpp#L52

So perhaps the way to go is to remove METIS/parMETIS after improving the usefulness of PETSc partitioning.  Should this happen with USE_METIS flags (like CVODE and VTK)?

jmpf:
Current plan for mesh partitioning types:

1. **METIS library** Remove completely, since it has been marked as  deprecated.  Remove from future Ubuntu packages and identify as dead in table of dependences.
2. **parMETIS library** either remove completely (as METIS) or allow to continue to support as a suggested package with USE_PARMETIS flags and a special portability build.
3. **PETSc with parMETIS** Have the current test for the existence of the interface fall back to a less sophisticated partitioner (MATPARTITIONINGAVERAGE or MATPARTITIONINGSQUARE?) rather than our dumb partitioner.
 
These 3 tasks could be treated independently.  However resolving parMETIS is important to #421 and PETSc/parMETIS is potentially most useful to any of our users who use high-performance computing.

jmpf:
In new PETSCs:
[...]

PR #437, opened 2025-09-23
#424 Finish deprecating METIS. Add parMETIS alternative.

jmpf:
1. Getting rid of METIS
2. Make parMETIS optional


heart/test/extended_bidomain/TestExtendedBidomainTissue.hpp
```diff
@@ -544,7 +544,12 @@ class TestExtendedBidomainTissue : public CxxTest::TestSuite
             HeartConfig::Instance()->SetMeshFileName("mesh/test/data/cube_136_elements");
 
             TrianglesMeshReader<3,3> mesh_reader("mesh/test/data/cube_136_elements");
+#ifndef CHASTE_SCOTCH_PARMETIS
             DistributedTetrahedralMesh<3,3> mesh;
+#else
+            // For libscotchparmetis turn off smart partitioning
```
mirams, line 550: Maybe make it clear in the test comments (which will be copied and pasted and used as a basis of simulations) that this isn't because scotch can't do it, but because it can't do it reproducibly?

(If I've understood correctly!)

mesh/src/common/DistributedTetrahedralMesh.cpp
```diff
@@ -113,8 +106,7 @@ void DistributedTetrahedralMesh<ELEMENT_DIM, SPACE_DIM>::ComputeMeshPartitioning
 {
     if (mPartitioning == DistributedTetrahedralMeshPartitionType::METIS_LIBRARY)
     {
-        WARNING("METIS partitioning is deprecated.  Switching to parMETIS");
-        mPartitioning = DistributedTetrahedralMeshPartitionType::PARMETIS_LIBRARY;
+        EXCEPTION("METIS partitioning is deprecated.");
```
mirams: "Please use PARMETIS or PT-SCOTCH instead" ?

heart/test/extended_bidomain/TestExtendedBidomainTissue.hpp
jmpf: My comments on these few tests weren't consistent.  I'm tidying them up.

mesh/src/common/DistributedTetrahedralMesh.cpp
jmpf: Fixing with suitable words.

mirams (approved):

Issue #435, opened 2025-09-14
VTK output for time-dependent mechanics problems


albertocorrias:
In cardiac Chaste, VTK output is implemented for electrophysiology problems where the mesh does not change over time. For electro-mechanics (EM) problems  like CardiacElectroMechanicsProblem (where the geometry changes over time), VTK output is not implemented. Instead, the output can be visualized through a software tool called [Cmgui](https://physiomeproject.org/software/opencmiss/cmgui). Such choice made sense at the time when the bulk of the EM code was written. However, Cmgui does not appear to be actively developed or maintained any longer (the [last available Debian package](https://physiomeproject.org/software/opencmiss/cmgui/download) is for Ubuntu 10.04 LTS) . It would be good to have the choice to visualize cardiac EM output using Paraview, i.e., outputting VTK files.

The continuum mechanics code currently includes a VtkNonlinearElasticitySolutionWriter. However, this class generates VTK output only once, displaying some typical mechanics results (pressures, displacements, etc) on the undeformed mesh when the equilibrium problem of continuum mechanics is solved. This class can't be used repeatedly in time-dependent problems, and also does not have ability to show other user-defined variables. 

Work in this ticket will include

1. Improving VtkNonlinearElasticitySolutionWriter (or writing a similar one), allowing writing multiple VTK files and using deformed or undeformed meshes for visualization. Perhaps such class should inherit form VTKMeshWriter. 
2. Allowing users to define custom output variables (e.g., voltage in EM problems) in VtkNonlinearElasticitySolutionWriter
3. Improve CardiacElectroMechanicsProblem by giving the user a choice of output between Cmgui and VTK (using the class at point 1).

PS. Although I have not written Chaste code in over a decade, I managed to write a working solution to this ticket on my machine to convince myself that it was possible to do so.

mirams:
Hi @albertocorrias, great to hear from you! Hope all is well.

I think we visualised for the PLoS 2013 paper, looks like a bit of custom code to write out the reference fibre directions (for a special plot of them, not normally needed), but nothing special for actual deformation, does the VtkMeshWriter 'just do it'?!:

https://chaste.github.io/docs/paper-tutorials/plos2013/electromechanics/ 

I'm 90% sure it was paraview to look at the output and make this video:

https://doi.org/10.1371/journal.pcbi.1002970.s006

albertocorrias:
Hi @mirams! Good to hear from you too!

[...]

Issue #436, opened 2025-09-23
Look into token-related build failures


fcooper8472:
E.g. here: https://github.com/Chaste/Chaste/pull/434

fcooper8472:
The failures are:

```
Error: Input required and not supplied: token
```

when checking out repos such as `memory-testing` and `profiling-gprof`. These repos require a token, which is stored in the Chaste repos secrets.

Secrets are *not* available to PRs running from forks, so this behaviour is currently expected when accepting an external contribution.

I suggest we migrate the steps involved in uploading the log files to a separate job, and make that job contingent on the pull request not coming from a fork.

This way it's business as usual for internal PRs, but external PRs will still run and correctly report success/failure in memory testing, but will not upload results to the website.

fcooper8472:
#441 splits each of the workflows that require token access into two jobs, the second of which shouldn't run if it's a fork.

This means we'll get:

- Genuine success/failure of profiling and memory testing on PRs from forks
- No data uploaded to the website

fcooper8472:
I've also noticed that CI on forks is failing in multiple places for other repos, such as Doxygen upload, updating tutorials etc, which don't make sense for forks to be running at all.

- [x] Add guards on those workflow runs so they only run on `Chaste/Chaste`

fcooper8472:
Final extra tweak needed: non-PR runs on forks also need to not access credentials.

PR #438, opened 2025-09-29
#435 Allow VTK visualization for Cardiac Electromechanics problems

albertocorrias:
Summary of the changes

-  [mesh/src/writer/VtkMeshWriter.cpp](https://github.com/Chaste/Chaste/compare/develop...albertocorrias:Chaste:develop#diff-f5cd0034bed9af1aa1aac060f41bbb48b965e0a165e81d8ee66341d750a2be1e): added a flag that, when false, will prevent the MakeVtkMesh() method from adding elements (aka faces) to the file. A test was added that shows that, without setting that flag to false, a second call to WriteFiles() on a modified mesh would cause a VTK error (and an assertion tripped).
 - Added a simple [mesh/src/writer/VtkDeformedMeshWriter.hpp](https://github.com/Chaste/Chaste/compare/develop...albertocorrias:Chaste:develop#diff-375bf779087fdaff5523505bd32cb5acfbf57a92a7d0b5b89286add3a8f5d9d6) class, inheriting from VtkMeshWriter and providing methods to modify the node coordinates and the file name to write to. A test was added.
- Through some refactoring, I made several methods of [continuum_mechanics/src/solver/VtkNonlinearElasticitySolutionWriter.cpp](https://github.com/Chaste/Chaste/compare/develop...albertocorrias:Chaste:develop#diff-8ed0bbf53c401f2526040d3c7ba6e5ea7c8199f11264f14093bda8527f85a9ab) available to be called by other classes.
 - Same re-factoring for [heart/src/postprocessing/VoltageInterpolaterOntoMechanicsMesh.cpp](https://github.com/Chaste/Chaste/compare/develop...albertocorrias:Chaste:develop#diff-0d466e7a249ac516f5d08e8e3e0bc8aa54de9b0d1d72d41fd7c54acee659f291)
 - Created a simple [heart/src/problem/CardiacElectroMechanicsVtkHandler.hpp](https://github.com/Chaste/Chaste/compare/develop...albertocorrias:Chaste:develop#diff-6bea5e27d045cce18dd4f3a1a0a5110d8072a6e8d2297c634f3d6f74ac17f13c) class that makes use of the re-factored classes mentioned above to orchestrate the generation of VTK output for electro-mechanics simulations.
- The Cardiac EM problem class was modified to use the CardiacElectroMechanicsVtkHandler when the user sets SetVisualizeWithVtk() from HeartConfig.
- The [heart/test/tutorials/TestCardiacElectroMechanicsTutorial.hpp](https://github.com/Chaste/Chaste/compare/develop...albertocorrias:Chaste:develop#diff-0ed2913c386d515c1efb022a2e9229c327f3d6c5c1e17643bb13291aeadc44cd) was modified to add VTK output and explanations for users were added


albertocorrias:
I can see the memtest failing, but the output says 100% tests passed. so I am not sure what to do with it. Could it be related to [PR441](https://github.com/Chaste/Chaste/pull/441)?

fcooper8472:
> I can see the memtest failing, but the output says 100% tests passed. so I am not sure what to do with it. Could it be related to [PR441](https://github.com/Chaste/Chaste/pull/441)?

@albertocorrias Yes, that's right, we've recently found that PRs opened from forks are not permitted to see the token required to check out the repos we use to keep reports.

You can ignore those two failing tests (but once #441 is merged, future PRs from forks should work as intended).

mirams:
Hi @albertocorrias, just to keep you up to date - @jmpf is having a look at your code and @fcooper8472 is testing fixes to the build system which should fix the above failures, so he will be in touch about updating to the latest develop soon if that is all working.

albertocorrias:
Thanks, @mirams and @jmpf ! Please do let me know if any (likely) improvement is needed

heart/src/postprocessing/VoltageInterpolaterOntoMechanicsMesh.hpp
```diff
@@ -39,36 +39,70 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#include "FineCoarseMeshPair.hpp"
 
 /**
- *  Very simple one-method class which can be used to convert the voltage from an electrics
+ *  A class which can be used to convert the voltage from an electrics
  *  (or electromechanics) simulation onto a coarser mechanics mesh, by interpolation. The
- *  class outputs a HDF5 file corresponding to nodes on the mechanics mesh, and converts it to
- *  CMGUI output.
+ *  class provides a method to interpolate values from the fine mesh 
+ *  onto the coarse mesh and the main meythod to output a HDF5 file corresponding to nodes
```
jmpf: meythod -> method

heart/src/problem/CardiacElectroMechanicsProblem.cpp
```diff
@@ -337,7 +337,13 @@ CardiacElectroMechanicsProblem<DIM,ELEC_PROB_DIM>::~CardiacElectroMechanicsProbl
     delete mpElectricsProblem;
     delete mpCardiacMechSolver;
     delete mpMeshPair;
-
+    
+#ifdef CHASTE_VTK
```
jmpf: Is all this logic necessary?  Would it be better to default to a null pointer so that delete becomes a no-op.

heart/test/mechanics/TestCardiacElectroMechanicsVtkHandler.hpp
```diff
@@ -0,0 +1,326 @@
...
+            // Check that it has the correct data
+            std::vector<double> voltage;
+            vtk_reader.GetPointData("V", voltage);
+            TS_ASSERT_EQUALS(voltage.size(),mechanics_mesh.GetNumNodes());
+            for (unsigned i=0u; i < voltage.size(); ++i)
+            {
+                TS_ASSERT_DELTA(fake_repl[i] , voltage[i], 1e-9);
+            }
+            
+            // No displacement expected
```
jmpf: Is it possible to a least fake something interesting here?  The problem is that when you run these two tests and visualise then all the voltages are 0 (in the first timestep), all the displacements are 0, and all deformation gradients are 0.  A user might believe this to be incorrect.

jmpf (commented):
I made some minor comments a while back.  Sorry - I forgot to press submit.

fcooper8472:
@albertocorrias we have recently resolved the issue of spurious test failures due to missing tokens, which is what is currently causing the red marks next to memory testing and profiling.

Could you please merge `develop` back into your branch to get the new workflows?

albertocorrias:
> @albertocorrias we have recently resolved the issue of spurious test failures due to missing tokens, which is what is currently causing the red marks next to memory testing and profiling.
> 
> Could you please merge `develop` back into your branch to get the new workflows?

Thanks @fcooper8472! I will do that, work on @jmpf's comments and push again.

heart/src/postprocessing/VoltageInterpolaterOntoMechanicsMesh.hpp
albertocorrias: Fixed in 4809d6f3a088e9d777e5a81d68e255c3d83379ef

heart/src/problem/CardiacElectroMechanicsProblem.cpp
albertocorrias: Some of it was not necessary, thanks. Changes  made in 4809d6f3a088e9d777e5a81d68e255c3d83379ef. However, the fact that VtkMeshWriter.hpp is completely enclosed in the CHASTE_VTK "if" statement, some logic in children classes must be there.

heart/test/mechanics/TestCardiacElectroMechanicsVtkHandler.hpp
albertocorrias: Great point. Thanks! 4809d6f3a088e9d777e5a81d68e255c3d83379ef made the test more robust by forcing a displacement and simulating a fake voltage gradient. Both are now clearly visible when, in ParaView, the user switches from t=0 (undeformed, constant voltages) to t=1 (clear deformation and voltage gradient).

jmpf: Thanks.  That last one is good because a user can now see that there's something going on in these fields.  It may be important for the future, because you/we might want to later add to pages such as https://chaste.github.io/docs/user-guides/visualisation-guides/paraview-for-cardiac/

jmpf:
Restarted tests...

heart/test/mechanics/TestCardiacElectroMechanicsVtkHandler.hpp
albertocorrias: There is a bit of explanation in the cardiac electromechanics tutorial, but it is quite short and without images.

I don't mind adding and updating that visualization guide page. It may need (1) more updated images (I am using Paraview 6 now, while the one shown is 3.4 and looking slightly different) and (2) adding a section on cardiac electromechanics. However, I am not quite sure where the source for those pages are stored and whether I can modify them.

jmpf: Updating visualisation guides is a task for another day: I think that https://github.com/Chaste/Chaste.github.io is a private repo so you'd need to get access or suggest changes.

jmpf:
@fcooper8472 and #436: Profiling failed with a token issue.

jmpf (approved):

mirams:
The tests are all passing, just some token trouble causing failure, so I will merge in

mirams:
Some memory test fixes here: https://github.com/Chaste/Chaste/pull/450 ...

PR #450, opened 2025-10-28
test memory fixes

mirams:
As a result of me prematurely merging in #438


albertocorrias:
Thanks, @mirams! I did not realize I should have deleted those vectors in the tests. I realized only now that the valgrind output is stored in the "memtest artifact".

mirams:
No problem, I had also forgotten, so @fcooper8472 has been enhancing the output so it's easier to see in future!

jmpf (approved):

Issue #453, opened 2025-10-30
New dependency versions: Boost 1.84 to 1.88


mirams:
Library that has a new versions: Boost 

Link to release notes: 

- 1.84 https://www.boost.org/releases/1.84.0/
- 1.85  https://www.boost.org/releases/1.85.0/
- 1.86  https://www.boost.org/releases/1.86.0/
- 1.87  https://www.boost.org/releases/1.87.0/
- 1.88  https://www.boost.org/releases/1.88.0/
- (1.89  https://www.boost.org/releases/1.89.0/  is now issue #465) 

**TODO:**
- [x]  Build a docker image containing the new dependency versions - see the [Portability Testing Guide](https://github.com/Chaste/developer_wiki/wiki/Portability-Testing-Guide) for how to do this.
- [x]  Create a branch off `develop` - you could name it, for example, `support-petsc-322`.
- [x]  Add the new docker image to the [portability workflow](https://github.com/Chaste/Chaste/blob/develop/.github/workflows/portability.yml) on the new branch.
- [x]  Create a PR with the new branch and see if it builds and passes all tests.
- [x]  Do any necessary fixes
  - [x] 1.84
  - [x] 1.85
  - [x] 1.86
  - [x] 1.87
  - [x] 1.88
- [x]  Add permanently to testing rotation i.e. either replace an old docker image in the [portability workflow](https://github.com/Chaste/Chaste/blob/develop/.github/workflows/portability.yml) with the new image, or append the new image to the existing list of portability images. For example, if there's an old dependency version that is being retired, the new docker image can probably replace the old one.
- [x]  Update the portability versions in the [Portability Testing Guide](https://github.com/Chaste/developer_wiki/wiki/Portability-Testing-Guide).
- [x]  Update the [supported dependency versions](https://chaste.github.io/docs/installguides/dependency-versions/) page.

jmpf:
I suggest that we defer supporting Boost 1.89 for the time-being and begin testing 1.88 (see PR #462)

jmpf:
See #449 - 25.10 Questing is now testing Boost 1.88.

jmpf:
> See [#449](https://github.com/Chaste/Chaste/issues/449) - 25.10 Questing is now testing Boost 1.88.

https://github.com/Chaste/dependency-modules/pull/92

PR #462, opened 2025-11-05
#453 Support for Boost up to 1.88

jmpf:
- [x] Fix 1.88 serialization library version
- [x] Find out when the serialization library version number changed
- [ ] ~Test on 1.89~


mirams (approved):
I guess we can fill in the gaps by checking output of portabilities if we have runners for all the intermediate ones... but probably fine like this

kwabenantim:
@jmpf I've come across a list of archive versions in [boostorg/serialization/basic_archive.cpp](https://github.com/boostorg/serialization/blob/develop/src/basic_archive.cpp) that seems to correspond with the versions we have in [TestArchivingHelperClasses.hpp](https://github.com/Chaste/Chaste/blob/develop/global/test/TestArchivingHelperClasses.hpp).

It doesn't say when every archive version was introduced, but it does say archive version 20 started with Boost 1.84 (the date on the comment is a bit odd but I suppose this probably narrows it down to versions 1.83, 1.84, and 1.85).

```cpp
// this should change if the capabilities are added to the library
// such that archives can be created which can't be read by previous
// versions of this library
// 1 - initial version
// 2 - made address tracking optional
// 3 - numerous changes - can't guarantee compatibility with previous versions
// 4 - Boost 1.34
//     added item_version to properly support versioning for collections 
// 5 - Boost 1.36
//     changed serialization of collections: adding version even for primitive
//     types caused backwards compatibility breaking change in 1.35
// 6 - Boost 1.41 17 Nov 2009
//     serializing collection sizes as std::size_t
// 7   Boost 1.42 2 Feb 2010
//     error - changed binary version to 16 bits w/o changing library version #
//     That is - binary archives are recorded with #6 even though they are
//     different from the previous versions.  This means that binary archives
//     created with versions 1.42 and 1.43 will have to be fixed with a special
//     program which fixes the library version # in the header
//     Boost 1.43 6 May 2010
//     no change
// 8 - Boost 1.44
//     separated version_type into library_version_type and class_version_type
//     changed version_type to be stored as 8 bits.
// 10- fixed base64 output/input.
// 11- not changes
// 12- improved serialization of collections
// 13- simplified visibility, removed Borland, removed pfto
// 14- improved visibility, refactor map/set
// 15- corrections to optional and collection loading
// 16- eliminated dependency on <codecvt> which is buggy in some libraries
//     and now officially deprecated in the standard
// 17- Boost 1.68 August 2018
// 18- addressed undefined behavior in archive constructors.
//     init() called from base wrote archive header before archive
//     was fully constructed.
// 19- Boost 1.76 April 2021
// 20- Boost 1.84 April 2021
```

jmpf:
There is a known issue with configuring Boost 1.89 in cmake.  My attempt to compile with 1.89 hit this same problem:
https://gitlab.com/OpenMW/openmw/-/issues/8754
I suggest we defer supporting it for the time-being.

PR #468, opened 2025-12-11
Remove boost system library, use header only

mirams:


jmpf (approved):
Looks like this is backward compatible for al the Boost we are now supporting.

PR #478, opened 2026-01-12
Adaptive timestep loop

### LINKED ISSUE #448: Cell Division Bug in Mutable Vertex Mesh leading to hanging/failing simulations.
**Describe the bug**
When running a mutable vertex mesh based proliferating cell simulation with a negative value of Lambda cells can possibly overlap their neighbours after cell division. This issue is caused by the addition of new nodes on an edge which is too small for division.

The offending code can be found on line 487 of mutable vertex mesh:

`if (norm_2(a_to_b) < 2.0 * mCellRearrangementRatio * mCellRearrangementThreshold)
            {
                WARNING("Edge is too small for normal division; putting node in the middle of a and b. There may be T1 swaps straight away.");
                ///\todo or should we move a and b apart, it may interfere with neighbouring edges? (see #1399 and #2401)
                intersection = position_a + 0.5 * a_to_b;
            }`

This can cause an immediate T1 transition that results in a cells edge overlapping into its neighbours body leading to an unrecoverable state leading to a silent crash without output. This does not happen 100% of the time when this section of code is executed, as through testing about 80% of simulations will be able to recover from this state.

**Proposed Change **
As described in the TODO of the code it may be better to increase the distance between node A and B such that it does not fall within this minimum tolerance. This should allow for correct cell division and prevent immediate T1 swapping.  An alternative would be to add a section of code to check if internal nodes are overlapping with their neighbours and kill the simulation with an error output to the user.

If cell vertices/nodes are to be moved apart a parameter will have to be selected to perform this move according to. Such that it is understandable and tuneable while also still being grounded in the physics of the mesh itself.

**Screenshots**
Example of failed simulation.

Image

BJackal:
This pull request relates to Issue #448 (https://github.com/Chaste/Chaste/issues/448).

In some rare instances, vertex mesh simulations would infinitely loop and never reached completion or exit due to any error. Initially thought to be a soft crash this was actually being caused due to the following lines of code in OffLatticeSimulation.cpp when adaptive time step was turned on (Starts on line 137).

```
        catch (StepSizeException& e)
        {
            // Detects if a node has travelled too far in a single time step
            if (mpNumericalMethod->HasAdaptiveTimestep())
            {
                // If adaptivity is switched on, revert node locations and choose a suitably smaller time step
                RevertToOldLocations(old_node_locations);
                present_time_step = std::min(e.GetSuggestedNewStep(), target_time_step - time_advanced_so_far);
            }
            else
            {
                // If adaptivity is switched off, terminate with an error
                EXCEPTION(e.what());
            }
        }
```

In cases where adaptive time step was turned on but a terminal error was never received, the catching of exceptions led to the incrementally decreasing time-step being called infinitely. As a way to fix this I suggest changing the code in the following way: 

```
catch (StepSizeException& e)
        {
            int adaptive_timer = 0;
            // Detects if a node has travelled too far in a single time step
            if (mpNumericalMethod->HasAdaptiveTimestep())
            {
                // If adaptivity is switched on, revert node locations and choose a suitably smaller time step
                RevertToOldLocations(old_node_locations);
                if (adaptive_timer < 5 ){
                present_time_step = std::min(e.GetSuggestedNewStep(), target_time_step - time_advanced_so_far);
                adaptive_timer += 1;
                } else {
                // If adaptivity is not improving the issue, terminate with an error to prevent infinite loop
                EXCEPTION(e.what());
                }
            }
            else
            {
                // If adaptivity is switched off, terminate with an error
                EXCEPTION(e.what());
            }
        }
```

 By adding the variable **adaptive_timer** and only allowing the loop to be called a pre-determined number of times (in this case 5) it prevents the possibility of the adaptive time step leading to infinite loops. 5 attempts at reducing the time-step feels like a reasonable number for attempting to save a simulation. This could even be made into a parameter the user could control.

### DIFF
--- a/cell_based/src/simulation/OffLatticeSimulation.cpp
+++ b/cell_based/src/simulation/OffLatticeSimulation.cpp
@@ -45,7 +45,8 @@ template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
 OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::OffLatticeSimulation(AbstractCellPopulation<ELEMENT_DIM,SPACE_DIM>& rCellPopulation,
                                                 bool deleteCellPopulationInDestructor,
                                                 bool initialiseCells)
-    : AbstractCellBasedSimulation<ELEMENT_DIM,SPACE_DIM>(rCellPopulation, deleteCellPopulationInDestructor, initialiseCells)
+    : AbstractCellBasedSimulation<ELEMENT_DIM,SPACE_DIM>(rCellPopulation, deleteCellPopulationInDestructor, initialiseCells),
+    mMaxAdaptiveTimeSteps(5)
 {
     if (!dynamic_cast<AbstractOffLatticeCellPopulation<ELEMENT_DIM,SPACE_DIM>*>(&rCellPopulation))
     {
@@ -95,6 +96,21 @@ const std::vector<boost::shared_ptr<AbstractForce<ELEMENT_DIM, SPACE_DIM> > >& O
     return mForceCollection;
 }
 
+// KARRIGAN
+template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
+void OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::SetMaxAdaptiveTimeStep(unsigned pMaxAdaptiveTimeStep)
+{
+    mMaxAdaptiveTimeSteps = pMaxAdaptiveTimeStep;
+}
+
+template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
+unsigned OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::GetMaxAdaptiveTimeStep() 
+{
+    return mMaxAdaptiveTimeSteps;
+}
+
+// KARRIGAN
+
 template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
 void OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::UpdateCellLocationsAndTopology()
 {
@@ -138,7 +154,7 @@ void OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::UpdateCellLocationsAndTopology
         {
             int adaptive_timer = 0;
             // Detects if a node has travelled too far in a single time step
-            if (mpNumericalMethod->HasAdaptiveTimestep() && adaptive_timer < 5 )
+            if (mpNumericalMethod->HasAdaptiveTimestep() && adaptive_timer < mMaxAdaptiveTimeSteps )
             {
                 // If adaptivity is switched on, revert node locations and choose a suitably smaller time step
                 RevertToOldLocations(old_node_locations);
--- a/cell_based/src/simulation/OffLatticeSimulation.cpp
+++ b/cell_based/src/simulation/OffLatticeSimulation.cpp
@@ -136,17 +136,25 @@ void OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::UpdateCellLocationsAndTopology
         }
         catch (StepSizeException& e)
         {
+            int adaptive_timer = 0;
             // Detects if a node has travelled too far in a single time step
             if (mpNumericalMethod->HasAdaptiveTimestep())
             {
                 // If adaptivity is switched on, revert node locations and choose a suitably smaller time step
                 RevertToOldLocations(old_node_locations);
+                if (adaptive_timer < 5 ){
                 present_time_step = std::min(e.GetSuggestedNewStep(), target_time_step - time_advanced_so_far);
+                adaptive_timer += 1;
+                } else {
+                // If adaptivity is not improving the issue, terminate with an error to prevent infinite loop
+                EXCEPTION(e.what());
+                }
             }
             else
             {
                 // If adaptivity is switched off, terminate with an error
                 EXCEPTION(e.what());
+                std::cout << "Reached here in Off Lattice Simulation 6\n" << std::endl;
             }
         }
     }
--- a/cell_based/src/simulation/OffLatticeSimulation.cpp
+++ b/cell_based/src/simulation/OffLatticeSimulation.cpp
@@ -154,7 +154,6 @@ void OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::UpdateCellLocationsAndTopology
             {
                 // If adaptivity is switched off, terminate with an error
                 EXCEPTION(e.what());
-                std::cout << "Reached here in Off Lattice Simulation 6\n" << std::endl;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


cell_based/src/simulation/OffLatticeSimulation.cpp
```diff
@@ -136,16 +136,18 @@ void OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::UpdateCellLocationsAndTopology
         }
         catch (StepSizeException& e)
         {
+            int adaptive_timer = 0;
             // Detects if a node has travelled too far in a single time step
-            if (mpNumericalMethod->HasAdaptiveTimestep())
+            if (mpNumericalMethod->HasAdaptiveTimestep() && adaptive_timer < 5 )
```
fcooper8472: @BJackal would it be possible to make this a bit more generic than having a magic number?

E.g. create an `unsigned mMaxAdaptivitySteps = 5;` in the OffLatticeSimulation, along with a getter and setter that you test?

fcooper8472 (changes requested):

cell_based/src/simulation/OffLatticeSimulation.cpp
BJackal: Thanks for the suggestions, I've now added them in.

cell_based/src/simulation/OffLatticeSimulation.cpp
```diff
@@ -95,6 +96,21 @@ const std::vector<boost::shared_ptr<AbstractForce<ELEMENT_DIM, SPACE_DIM> > >& O
     return mForceCollection;
 }
 
+// KARRIGAN
+template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
+void OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::SetMaxAdaptiveTimeStep(unsigned pMaxAdaptiveTimeStep)
```
fcooper8472: `p` prefix is reserved for pointers. Parameter can just be `maxAdaptiveTimeStep`

cell_based/src/simulation/OffLatticeSimulation.hpp
```diff
@@ -103,6 +103,9 @@ class OffLatticeSimulation : public AbstractCellBasedSimulation<ELEMENT_DIM,SPAC
     /** The numerical method to use in this simulation. Defaults to the explicit forward Euler method. */
     boost::shared_ptr<AbstractNumericalMethod<ELEMENT_DIM, SPACE_DIM> > mpNumericalMethod;
 
+    /** Variable for the allowed number of adaptive time step changes before outputting underlying exception. */
+    int mMaxAdaptiveTimeSteps;
```
fcooper8472: This should be `unsigned` rather than `int`

cell_based/src/simulation/OffLatticeSimulation.hpp
```diff
@@ -186,6 +189,18 @@ class OffLatticeSimulation : public AbstractCellBasedSimulation<ELEMENT_DIM,SPAC
      */
     const boost::shared_ptr<AbstractNumericalMethod<ELEMENT_DIM, SPACE_DIM> > GetNumericalMethod() const;
 
+    /**
+     * Set the maximum number of adaptive timesteps that can be attempted before outputting the underlying exception.
+     *
+     * @param pMaxAdaptiveTimeStep max adaptive timestep parameter
+     */
+    void SetMaxAdaptiveTimeStep(unsigned pMaxAdaptiveTimeStep);
```
fcooper8472: Should be `void SetMaxAdaptiveTimeStep(const unsigned pMaxAdaptiveTimeStep);` (const parameter)

cell_based/src/simulation/OffLatticeSimulation.hpp
```diff
@@ -186,6 +189,18 @@ class OffLatticeSimulation : public AbstractCellBasedSimulation<ELEMENT_DIM,SPAC
...
+
+    /**
+     * @return the maximum allowed number of attempted timestep reductions.
+     */
+    unsigned GetMaxAdaptiveTimeStep();
```
fcooper8472: Should be `unsigned GetMaxAdaptiveTimeStep() const;` (const member function, i.e. you're signaling that this method does not modify the `OffLatticeSimulation` object.

cell_based/test/simulation/TestOffLatticeSimulation.hpp
```diff
@@ -1375,6 +1375,7 @@ class TestOffLatticeSimulation : public AbstractCellBasedWithTimingsTestSuite
         simulator.SetOutputDirectory("TestOffLatticeSimulationWithAdaptivity");
         simulator.SetEndTime(5.0);
         simulator.SetDt(0.1);
+        simulator.SetMaxAdaptiveTimeStep(3);
```
fcooper8472: Once you update the member to `unsigned` then technically this should be `3u` (the `u` being the suffix for unsigned, which then prevents an implicit conversion form `3` (an int).

cell_based/test/simulation/TestOffLatticeSimulation.hpp
```diff
@@ -1393,6 +1394,7 @@ class TestOffLatticeSimulation : public AbstractCellBasedWithTimingsTestSuite
         simulator.Solve();
 
         TS_ASSERT_EQUALS(simulator.rGetCellPopulation().GetNumRealCells(), 14u);
+        TS_ASSERT_EQUALS(simulator.GetMaxAdaptiveTimeStep(),3);
```
fcooper8472: Ditto, change this to `3u` (and add a space after the comma)

cell_based/src/simulation/OffLatticeSimulation.cpp
```diff
@@ -45,7 +45,8 @@ template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
 OffLatticeSimulation<ELEMENT_DIM,SPACE_DIM>::OffLatticeSimulation(AbstractCellPopulation<ELEMENT_DIM,SPACE_DIM>& rCellPopulation,
                                                 bool deleteCellPopulationInDestructor,
                                                 bool initialiseCells)
-    : AbstractCellBasedSimulation<ELEMENT_DIM,SPACE_DIM>(rCellPopulation, deleteCellPopulationInDestructor, initialiseCells)
+    : AbstractCellBasedSimulation<ELEMENT_DIM,SPACE_DIM>(rCellPopulation, deleteCellPopulationInDestructor, initialiseCells),
+    mMaxAdaptiveTimeSteps(5)
```
fcooper8472: This is fine, but, this technically means only if this specific constructor is used does the value 5 get set as a default. Any other constructors would leave the value unset.

Since C++11 you have been able to specify an "in-class initialiser", in the header file, which in this case is probably preferred, as it will set a global default for the class.

So, remove from the init list here, and in the header:

```cpp
unsigned mMaxAdaptiveTimeSteps = 5;
```

cell_based/src/simulation/OffLatticeSimulation.hpp
fcooper8472: (and see other comment: `unsigned mMaxAdaptiveTimeSteps = 5;` to set global default value.

fcooper8472 (changes requested):
Only v minor things to update, but this look good other than those small tweaks!

fcooper8472 (approved):
Looks good!

@BJackal feel free to merge as soon as you get a full set of green ticks on the CI

PR #501, opened 2026-04-10
Clarify nvpl warnings are safe to ignore for affected CMake users

### LINKED ISSUE #477: Warnings when finding BLAS/LAPACK from CMake 4.1+
When we find SUNDIALS, CMake internally finds BLAS and LAPACK.

From CMake 4.1, they added support for the [Nvidia Performance Libraries BLAS implementation](https://developer.nvidia.com/nvpl).

This has introduced a call to

```cmake
    find_package(nvpl)
```

This is not installed with `chaste-dependencies`, and so emits the following warning:

```
CMake Warning at /home/fcooper/.cache/JetBrains/RemoteDev/dist/4537479f9877a_CLion-2025.3.1/bin/cmake/linux/x64/share/cmake-4.1/Modules/FindLAPACK.cmake:740 (find_package):
  By not providing "Findnvpl.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "nvpl", but
  CMake did not find one.

  Could not find a package configuration file provided by "nvpl" with any of
  the following names:

    nvplConfig.cmake
    nvpl-config.cmake

  Add the installation prefix of "nvpl" to CMAKE_PREFIX_PATH or set
  "nvpl_DIR" to a directory containing one of the above files.  If "nvpl"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  cmake/Modules/FindSUNDIALS.cmake:159 (find_package)
  CMakeLists.txt:554 (find_package)
```

This is big and red and printed 4 times, and even though it ultimately does find an acceptable BLAS, we really need to work out the best way to silence this warning.

It's not urgent unless Ubuntu 26.04 ships with that version of CMake, but often IDEs bundle a newer version than the system so we should sort this out fairly quickly.

fcooper8472:
See #477

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1140,6 +1140,20 @@ foreach (component ${Chaste_COMPONENTS})
 endforeach (component ${Chaste_COMPONENTS})
 
 
+# CMake 4.1.0-4.1.4 and 4.2.0-4.2.1 emit spurious "Could not find a package configuration file provided by nvpl"
+# warnings while searching for BLAS/LAPACK (triggered when finding PETSc and SUNDIALS). These warnings are harmless and
+# can safely be ignored.
+# The issue was fixed in CMake 4.1.5 and 4.2.2 respectively:
+# https://gitlab.kitware.com/cmake/cmake/-/work_items/27451
+# https://github.com/Chaste/Chaste/issues/477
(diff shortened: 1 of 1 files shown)


jmpf (approved):
Okay - as long as users read this.

kwabenantim (approved):
This looks like a sound strategy to me.

PR #503, opened 2026-04-20
Support VTK 9.4 and 9.5

### LINKED ISSUE #458: New dependency version: VTK versions 9.4 and 9.5
Library that has a new version: VTK

Link to release notes: 
 * 9.4 https://docs.vtk.org/en/latest/release_details/9.4.html
 * 9.5 https://docs.vtk.org/en/latest/release_details/9.5.html

**TODO:**
  - [x] 9.4
  - [x] 9.5
- [x]  Add permanently to testing rotation i.e. either replace an old docker image in the [portability workflow](https://github.com/Chaste/Chaste/blob/develop/.github/workflows/portability.yml) with the new image, or append the new image to the existing list of portability images. For example, if there's an old dependency version that is being retired, the new docker image can probably replace the old one.
- [x]  Update the [supported dependency versions](https://chaste.github.io/docs/installguides/dependency-versions/) page.

fcooper8472:
See #458

### DIFF
--- a/lung/src/airway_generation/AirwayGenerator.cpp
+++ b/lung/src/airway_generation/AirwayGenerator.cpp
@@ -54,15 +54,14 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #if ((VTK_MAJOR_VERSION >= 5 && VTK_MINOR_VERSION >= 6) || VTK_MAJOR_VERSION >= 6)
 
-#include "vtkGeometryFilter.h"
 #include "vtkDoubleArray.h"
+#include <vtkClipPolyData.h>
 #include "vtkPlane.h"
 #include "vtkPoints.h"
 #include "vtkPointData.h"
 #include "vtkPolyVertex.h"
 #include "vtkMassProperties.h"
 #include "vtkMath.h"
-#include "vtkTableBasedClipDataSet.h"
 #include "vtkCellArray.h"
 #include "vtkLine.h"
 #include "vtkUnsignedIntArray.h"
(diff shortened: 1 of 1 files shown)


lung/src/airway_generation/AirwayGenerator.cpp
```diff
@@ -235,26 +234,19 @@ vtkSmartPointer<vtkPolyData> AirwayGenerator::SplitPointCloud(vtkSmartPointer<vt
...
 {
-    vtkSmartPointer<vtkPlane> clip_plane = vtkSmartPointer<vtkPlane>::New();
+    auto clip_plane = vtkSmartPointer<vtkPlane>::New();
     clip_plane->SetNormal(rNormal);
     clip_plane->SetOrigin(rOrigin);
 
-    vtkSmartPointer<vtkTableBasedClipDataSet > clipper = vtkSmartPointer<vtkTableBasedClipDataSet >::New();
-#if VTK_MAJOR_VERSION >= 6
+    auto clipper = vtkSmartPointer<vtkClipPolyData>::New();
     clipper->SetInputData(pointCloud);
```
jmpf, line 242: Good to remove unsupported version code

lung/src/airway_generation/AirwayGenerator.cpp
```diff
@@ -235,26 +234,19 @@ vtkSmartPointer<vtkPolyData> AirwayGenerator::SplitPointCloud(vtkSmartPointer<vt
...
     clipper->SetInsideOut(insideOut);
+    clipper->GenerateClippedOutputOff();
     clipper->Update();
 
-    vtkSmartPointer<vtkGeometryFilter> filter = vtkSmartPointer<vtkGeometryFilter>::New();
-    filter->SetInputConnection(clipper->GetOutputPort());
-    filter->Update();
-
-    return filter->GetOutput();
+    return clipper->GetOutput();
```
jmpf, line 249: Works on all previous versions from 6 up?

jmpf (commented):
Looks good, subject to back-compatibility

lung/src/airway_generation/AirwayGenerator.cpp
fcooper8472: @jmpf not necessarily a high priority, but there's a lot of this guarding in lung, and we haven't supported VTK < 7 for quite some time.

In fact, it's sort of a myth these days that VTK is optional, and perhaps we should just remove all the guarding and simplify the codebase..!

lung/src/airway_generation/AirwayGenerator.cpp
fcooper8472: It passes all current portability tests, and the oldest we check is Ubuntu 22.04 which ships VTK 7.1. So I haven't checked 6.x, but we make no guarantee that those version still work with Chaste.

fcooper8472:
@jmpf and @kwabenantim this should now pass all tests, and is ready to merge if you're happy with the VTK changes.

jmpf (approved):

PR #514, opened 2026-04-30
Modernize: replace BOOST_FOREACH with range-based for loops

mirams:
Replaces all BOOST_FOREACH(decl, range) usages with standard range-based for (decl : range) loops and removes the #include <boost/foreach.hpp> directives. Removes BOOST_FOREACH from .clang-format ForEachMacros list.

I like this because it is removing a header use, requires less magic knowledge of obscure boost macros.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -44,7 +44,7 @@ Cpp11BracedListStyle: false
 DerivePointerAlignment: false
 DisableFormat:   false
 ExperimentalAutoDetectBinPacking: false
-ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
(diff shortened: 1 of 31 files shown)


jmosborne:
For what its worth I'm happy to review.

mirams:
OK, @jmosborne, @jmpf, @fcooper8472 this is the first of a few copilot efforts at modernising our C++ to use up to C++17 features. This one I like because it is simpler in terms of using core language, removing a boost header use, less of a magical macro. So a nice easy one to include I think.

fcooper8472 (approved):

fcooper8472:
@mirams looks good to me!

PR #515, opened 2026-04-30
Modernize: replace boost::shared_ptr with std::shared_ptr

mirams:
### Summary
Replaces boost::shared_ptr, boost::make_shared, boost::weak_ptr, boost::enable_shared_from_this, and boost pointer casts with their std:: equivalents. Updates corresponding #include directives to use <memory> instead of boost headers.

**To Do**
- [x] Replace `boost::shared_ptr` and related smart pointers with corresponding `std::shared_ptr equivalents`.
- [x] Support the changes in chaste-codegen:
        https://github.com/ModellingWebLab/chaste-codegen/pull/297.
- [ ] Support the changes in the `CellBasedComparison2017` project.
- [ ] { Support the changes in any other related projects required to pass CI. }
- [ ] Add a "breaking changes" note to release notes: 
       [Chaste.github.io/site/content/docs/release-notes/changes-since-last-release.md](https://github.com/Chaste/Chaste.github.io/blob/main/site/content/docs/release-notes/changes-since-last-release.md)


mirams:
OK, so https://github.com/ModellingWebLab/chaste-codegen would need updating for this too...

kwabenantim:
**PyChaste related notes**
* In `pychaste/dynamic/config.yaml`, `smart_ptr_type: boost::shared_ptr` will need updating.
* Also needs updating `pychaste/dynamic/templates/CellsGeneratorCustomTemplate.py` (this particular edit could be made unnecessary by https://github.com/Chaste/cppwg/issues/132).

kwabenantim:
`Cell` never actually uses [`shared_from_this`](https://en.cppreference.com/cpp/memory/enable_shared_from_this) which hides a latent problem that comes up when something else (like pybind11) tries to use it. What causes the problem is that `Cell`'s constructor does `CellPtr(this, null_deleter())` before any owner exists, which makes `shared_from_this()` return a non-owning pointer later.

This didn't happen with `boost::shared_ptr` because pybind11 doesn't understand it very deeply and doesn't try to do anything clever with it.

This is addressed in 2ef5391.

kwabenantim:
> OK, so https://github.com/ModellingWebLab/chaste-codegen would need updating for this too...

Chaste currently pins `chaste_codegen>=0.10.0, <0.11.0`.  We could:
- Temporarily install the codegen on this branch directly from codegen's own `std::shared_ptr` [branch](https://github.com/ModellingWebLab/chaste-codegen/pull/297).
- Release the `std::shared_ptr` codegen changes first as `0.11.0`
- Bump this branch up to `chaste_codegen>=0.11.0, <0.12.0`
-  Merge when testing passes.

kwabenantim:
This also affects the [CellBasedComparison2017](https://github.com/Chaste/Chaste/actions/runs/33627955220/job/100240091755?pr=515) project, and possibly other downstream projects so should probably add a note to the breaking changes section in release notes as part of the work related to this PR.

PR #519, opened 2026-05-01
Update base image for regenerating archive

### LINKED ISSUE #518: Generate Crypt Archive broken
The method to generate a crypt archive 

[https://github.com/Chaste/developer_wiki/wiki/Regenerate-Crypt-archive](https://github.com/Chaste/developer_wiki/wiki/Regenerate-Crypt-archive)

is broken as its ctest is too old. The error is 

`[1:40 PM]Run source init.sh
CMake Error at CMakeLists.txt:34 (cmake_minimum_required):
  CMake 3.22.1 or higher is required.  You are running version 3.16.3
-- Configuring incomplete, errors occurred!`

fcooper8472:
See #518 

@kwabenantim you're definitely the best person to review this PR. Can you check whether this is likely to work, and advise if there's a better solution?

I think the oldest supported boost will always be on an Ubuntu LTS (unless something strange happens), so it will very likely always be possible to use a portability system image for this.

### DIFF
--- a/.github/workflows/regenerate-archive.yml
+++ b/.github/workflows/regenerate-archive.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
 
     container:
-      image: chaste/runner:oldest-boost
+      image: chaste/runner:portability-jammy
       env:
         RUNNER_OFF: 1
       volumes:


kwabenantim:
There are a couple of [special docker tags](https://github.com/Chaste/developer_wiki/wiki/Portability-Testing-Guide#adding-extra-image-tags) that can be pointed to a different image when we move versions up: `oldest-boost` and `coverage-runner`. This was useful so we could move to a different image without editing the respective workflows.

Alternatively, we could get rid of the pointer tags and use the image names directly if that makes things clearer @fcooper8472.

fcooper8472:
@kwabenantim good question! My hunch is it's simpler to use the image names directly, as that will get rid of a layer of obfuscation.

What does `coverage-runner` point to?

fcooper8472:
This change has fixed the workflow: https://github.com/Chaste/Chaste/actions/runs/25218920096

kwabenantim:
> My hunch is it's simpler to use the image names directly, as that will get rid of a layer of obfuscation.

I think that's a good call.

> What does coverage-runner point to?

[coverage-runner](https://hub.docker.com/layers/chaste/runner/coverage-runner/images/sha256-7adfa4d9974c220c8bbc6e258b7052e4242b8165473ff50087b6d6956751a61d) currently points to a previous version of `portability-01` (which has since moved up to new dependency versions).

kwabenantim (approved):
Looks good to me.

jmosborne:
The new setup worked a treat!

PR #534, opened 2026-05-02
Support PETSc 3.25

### LINKED ISSUE #522: New dependency version: PETSc 3.25
Library that has a new version: PETSc 3.25

Link to release notes: https://petsc.org/release/changes/325/

**TODO:**
- [x]  Add permanently to testing rotation i.e. either replace an old docker image in the [portability workflow](https://github.com/Chaste/Chaste/blob/develop/.github/workflows/portability.yml) with the new image, or append the new image to the existing list of portability images. For example, if there's an old dependency version that is being retired, the new docker image can probably replace the old one.
- [x]  Update the [supported dependency versions](https://chaste.github.io/docs/installguides/dependency-versions/) page.

fcooper8472:


jmpf (approved):
Note from  https://chaste.github.io/docs/installguides/dependency-versions/ that PETSc < 3.15 are now officially unsupported by us.  Both the old interface and the new interface worked in version 3.14 meaning that we don't have to special case the old interface anywhere in our code.

PR #542, opened 2026-05-19
Support HDF5 2.x

### LINKED ISSUE #521: New dependency version: HDF5 2.0, 2.1
Library that has a new version: HDF5 2.0, 2.1

Link to release notes: https://github.com/HDFGroup/hdf5/releases

2.0 is a major release of HDF5, including a [new file format](https://github.com/HDFGroup/hdf5/blob/2.0.0/release_docs/CHANGELOG.md#fileformat):

> This new format is only used when the HDF5 library version bounds lower bound is set to 2.0 or later.

There are also API changes.

**TODO:**
- [x] (For this issue:) add a backward compatibility test
- [x]  Add permanently to testing rotation i.e. either replace an old docker image in the [portability workflow](https://github.com/Chaste/Chaste/blob/develop/.github/workflows/portability.yml) with the new image, or append the new image to the existing list of portability images. For example, if there's an old dependency version that is being retired, the new docker image can probably replace the old one. (https://github.com/Chaste/dependency-modules/issues/128)
- [x]  Update the [supported dependency versions](https://chaste.github.io/docs/installguides/dependency-versions/) page. (https://github.com/Chaste/Chaste.github.io/pull/210)

kwabenantim:
Supports #521


kwabenantim:
The `HDF5_IS_PARALLEL` CMake flag has been renamed to `HDF5_PROVIDES_PARALLEL` in HDF5 2.0+

CMakeLists.txt
```diff
@@ -497,7 +497,14 @@ message(STATUS "HDF5 version: ${HDF5_VERSION}")
 target_link_libraries(Chaste_COMMON_DEPS INTERFACE HDF5::HDF5)
 
 # Chaste requires a parallel HDF5 library
-if (NOT HDF5_IS_PARALLEL)
+set(HDF5_HAS_PARALLEL FALSE)
```
fcooper8472:
```suggestion
set(Chaste_HDF5_HAS_PARALLEL FALSE)
```

@kwabenantim this looks fine, but could we prepend this with `Chaste_` so it's clear that this is our variable and not something imported when we find HDF5?

kwabenantim: Thanks @fcooper8472. This  is now fixed in 4da4682

mirams (approved):

.github/workflows/portability.yml
```diff
@@ -39,6 +39,8 @@ jobs:
             extra-build-args: ""
           - image: portability-resolute
             extra-build-args: ""
+          - image: portability-dev
```
fcooper8472: Need to remove this before merging - @kwabenantim does one of the numbered portabilities include HDF5 2.x?

fcooper8472 (dismissed):

fcooper8472 (approved):

PR #568, opened 2026-07-09
Petsc gpu support

### LINKED ISSUE #528: Investigate PETSc GPU support
### Info

> PETSc uses a single source programming model where solver back-ends are selected as runtime options and configuration options with no changes to the API.

https://petsc.org/main/overview/gpu_roadmap/#doc-gpu-roadmap
Supports CUDA & OpenCL for vectors and matrices

Also supports some numerical routines via openmp if built with `--with-openmp` and `--with-openmp-kernels`

Compile with `--with-cuda` to enable CUDA acceleration. (Looks like this can be used with openmp as well)

### PETSc setup 
https://petsc.org/main/install/install/#doc-config-accel

### Usage
https://petsc.org/main/faq/#doc-faq-gpuhowto

### Setup & Usage

GPU acceleration requires:
- PETSc to be built from source with one of the relevant flags (e.g. `./configure --with-cuda`)
- A gpu-aware MPI installation (if petsc `--download-openmpi/mpich` option is used, petsc will automatically configure mpi to be gpu-aware)
- A valid installation of the relevant gpu toolkit, e.g. CUDA, HIP

GPU-aware petsc supports vector and matrix operations. The user must pass `-vec_type` and `-mat_type` options to the executable at runtime, followed by a vector/matrix format. Various options for vector/matrix format exist, e.g. `mpicuda`, `mpiaijcusparse` If gpu-aware types are selected, PETSc should offload them.

If you require finer control over the types, i.e. you want to use more than one vector or matrix type in a program, the type must be set in the source code rather than at runtime.

### Results

- Some tests would pass using some vector/matrix types. For vectors, `mpicuda` was most stable, and for matrices `mpiaijcusparse` was most stable. Other types would typically cause errors.
- Even with these types, some tests would still error out. This could be due to unsupported operations or multiple vector/matrix types being required.
- In all test cases that did work, the GPU was slower than the CPU (GTX1070 vs Intel® Core™ i5-6600K)
- Some test cases would complete without error but produce incorrect results or at least results outside of the test tolerances

### Discussion

It would take a fairly significant development effort to adapt the chaste source code to be truly compatible with the PETSc GPU infrastructure. It is not clear that there would be a significant performance improvement as a result. Various PETSc operations cause data to be transferred back and forth between the GPU and host, algorithms would need to be re-written to avoid these. In addition, separate code paths would be required for any executables requiring more than one vector/matrix type. Given chaste is a library, we can't predict what combination of library objects end users will want to use, so realistically we would need to do this everywhere.

### Conclusions
- Set up wouldn't be too complicated using `dependency-modules`, although the CMake infrastructure would need extending to support passing the vector/matrix types to the executables if runtime configuration was desired.
- Probably not worth the development effort at the moment.
- Potentially worth considering if we have a particular need for it at some point, i.e. one slow bit of code.

MILeach:
Adds support for CUDA-aware PETSc builds. Tested with GTX1070, CUDA 11.3, PETSc 3.19.6

## Usage

- Point chaste to a CUDA enabled PETSc installation
- Configure chaste with the `Chaste_ENABLE_PETSC_CUDA` flag set to `ON`
- CUDA types will be used by default if this flag is set

### DIFF
--- a/cmake/Modules/FindPETSc.cmake
+++ b/cmake/Modules/FindPETSc.cmake
@@ -131,6 +131,21 @@ if(PC_PETSc_VERSION)
     unset(_petsc_version_patch)
 endif()
 
+# Detect whether the found PETSc was built with CUDA support, by inspecting petscconf.h
+set(PETSc_HAVE_CUDA FALSE)
+foreach(_petsc_include_dir ${PETSc_INCLUDE_DIRS})
+    if(EXISTS "${_petsc_include_dir}/petscconf.h")
+        file(STRINGS "${_petsc_include_dir}/petscconf.h" _petsc_have_cuda_line REGEX "^#define PETSC_HAVE_CUDA 1")
+        if(_petsc_have_cuda_line)
+            set(PETSc_HAVE_CUDA TRUE)
+        endif()
+        unset(_petsc_have_cuda_line)
+    endif()
+endforeach()
+unset(_petsc_include_dir)
+set(PETSc_HAVE_CUDA ${PETSc_HAVE_CUDA} CACHE BOOL "Whether the found PETSc was built with CUDA support")
+mark_as_advanced(PETSc_HAVE_CUDA)
+
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(PETSc
     REQUIRED_VARS PETSc_FOUND PETSc_INCLUDE_DIRS PETSc_LIBRARIES
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,7 @@ endif ()
 
 option (Chaste_USE_PETSC_PARMETIS "Prefer to compile Chaste with PARMETIS library used by PETSc" ON)
 option (Chaste_USE_PETSC_HDF5 "Prefer to compile Chaste with HDF5 library used by PETSc" ON)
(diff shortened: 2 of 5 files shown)


MILeach:
@MILeach to update documentation to describe how to use this. Consider adding a tutorial

cmake/Modules/FindPETSc.cmake
```diff
@@ -131,6 +131,21 @@ if(PC_PETSc_VERSION)
     unset(_petsc_version_patch)
 endif()
 
+# Detect whether the found PETSc was built with CUDA support, by inspecting petscconf.h
+set(PETSc_HAVE_CUDA FALSE)
+foreach(_petsc_include_dir ${PETSc_INCLUDE_DIRS})
+    if(EXISTS "${_petsc_include_dir}/petscconf.h")
+        file(STRINGS "${_petsc_include_dir}/petscconf.h" _petsc_have_cuda_line REGEX "^#define PETSC_HAVE_CUDA 1")
+        if(_petsc_have_cuda_line)
+            set(PETSc_HAVE_CUDA TRUE)
```
jmpf: Can you replace all of this with PETSC_HAVE_CUDA ?

In /usr/lib/petscdir/petsc3.19/x86_64-linux-gnu-real/include/petscsys.h
```
#if defined(PETSC_HAVE_CUDA)
```

Issue #324, opened 2024-10-03
Simplify monitoring of VTK libexpat-related backports


fcooper8472:
Related to #249 

Currently we have a sophisticated workflow that runs the Continuous test suite on Ubuntu 22.04 without downgrading the version of libexpat.

This is therefore expected to fail until VTK is patched.

However, we would prefer to modify the chaste-dependencies package to pull in the patched VTK instead of asking our users to downgrade libexpat, as Chaste will then continue to work "out of the box" on Ubuntu 22.04, which we will support until April 2027.

Once this change is made, the current monitoring will no longer work. I propose a radically simpler approach:

- Run on Ubuntu 22.04 and 24.04
- Install VTK and check which version we end up with

If it doesn't match the version that exists at time of writing this issue, then they have updated it and we can go in an check whether it fixes the libexpat problem.

fcooper8472:
Current versions

Noble 24.04:
- `libvtk9-dev` gives `9.1.0+really9.1.0+dfsg2-7.1build3`

Jammy 22.04:
- `libvtk9-dev` gives `9.1.0+really9.1.0+dfsg2-3build1`
- `libvtk7-dev` gives `7.1.1+dfsg2-10.1build1`

PR #582, opened 2026-07-23
Update workflows from ubuntu 2204 to ubuntu 2604 runners

### LINKED ISSUE #581: Update workflows from ubuntu-22.04 to ubuntu-26.04 runners
ubuntu-26.04 is now available on GitHub Actions. Update all workflows currently using ubuntu-22.04 runners to use ubuntu-26.04.

fcooper8472:
See #581.

Moves CI off the ubuntu-22.04 GitHub runner (no longer needed now that 22.04 support is dropped), keeping the existing 24.04 jobs intact.

### Runner updates (native GitHub runners)
- `ubuntu-22.04` → `ubuntu-26.04` and apt suite `jammy` → `resolute` in: `ubuntu-tests`, `ubuntu-basic-compatibility`, `memory-testing`, `profiling-gprof`, `project-tests`.
- 24.04/noble jobs left unchanged.

### Retired VTK version monitoring
- Deleted `vtk-version-monitoring.yml`. It existed to detect when the libexpat/VTK backport situation changed (see #324, now closed), but there is no realistic prospect of the fix being backported to 24.04 at this point.

### Compilers workflow
- Renamed `ubuntu-2204-compilers.yml` → `ubuntu-2604-compilers.yml`, moved to `ubuntu-26.04`/`resolute`, and updated the matrix to 26.04 toolchains: gcc 15/16, clang 21/22, plus icx. gcc-16 is `apt`-installed (not pre-installed on the runner); the rest use the runner's pre-installed compilers.

### Action version bumps (repo-wide)
- Updated GitHub Actions to their latest releases: `checkout` → v7, `upload-artifact` → v7, `setup-python` → v7, `setup-go` → v7. (`download-artifact@v8`, `cache@v6`, `setup-miniconda@v4`, `docker/github-builder@v1` already latest.)


fcooper8472:
~@kwabenantim any ideas why that pychaste conda test is failing? It doesn't seem related to this PR, but I'm not certain enough to ignore it without your input.~

These are removed by https://github.com/Chaste/Chaste/pull/586

fcooper8472:
~The coverage failures look spurious. There is no indication if you click through to codecov that coverage has reduced at all.~

Coverage correct on next run. Spurious.

fcooper8472:
@kwabenantim https://github.com/Chaste/Chaste/pull/582/commits/04bdf17b27a856715c19ff9eb95a6a6ae2587380 addresses what I believe was a real memory leak in the `CvodeContextManager`. Could you take a look and see if you're happy with the fix?

`Instance()` held the per-thread `SUNContext` in a `thread_local unique_ptr`. Thread-local destructors run at the start of `exit()`, before the destructors of statics in shared libraries. `CellCycleModelOdeSolver::mpInstance` is one of those statics and still holds a `CvodeAdaptor`, whose destructor calls `CVodeFree()`. So the context was freed first and then read:

```
Invalid read of size 4
   at SUNContext_PeekLastError (libsundials_core.so.7.1.1)
   by SUNNonlinSolFree_Newton / CVodeFree
   by CvodeAdaptor::FreeCvodeMemory() (CvodeAdaptor.cpp:390)
   by __cxa_finalize / _dl_fini
 Address is inside a block free'd by
      SUNContext_Free / ~CvodeContextManager() / __call_tls_dtors
```

Fix: `Instance()` returns `std::shared_ptr<CvodeContextManager>`, and the two classes owning long-lived Sundials objects (`CvodeAdaptor`, and `AbstractCvodeSystem`) keep a copy. Members are destroyed after the destructor body, so `CVodeFree()` always sees a live context and the last release frees it.

Why it was never caught: the memtest job ran on 22.04, which ships Sundials 5.8.0, and the whole class sits behind `#if CHASTE_SUNDIALS_VERSION >= 60000`. Sundials 6 (24.04) has the class and the bug but memtest never ran on 24.04.

kwabenantim:
> https://github.com/Chaste/Chaste/commit/04bdf17b27a856715c19ff9eb95a6a6ae2587380 addresses what I believe was a real memory leak in the CvodeContextManager. Could you take a look and see if you're happy with the fix?

This looks good to me. Switching `CvodeContextManager` to a shared pointer  and keeping it alive in `CvodeAdaptor` and `AbstractCvodeSystem` is the safe thing to do.

Not directly related, but we only [officially support](https://chaste.github.io/docs/installguides/dependency-versions/) Sundials 6+ now so can probably drop the guards at some point.

.github/workflows/coverage.yml
kwabenantim, line 19:
```suggestion
      image: chaste/runner:portability-resolute
```

kwabenantim: Currently, [coverage-runner](https://hub.docker.com/layers/chaste/runner/coverage-runner/images/sha256-7adfa4d9974c220c8bbc6e258b7052e4242b8165473ff50087b6d6956751a61d) seems to point to an old Ubuntu 22.04 image. It might be a good idea to retire the extra `coverage-runner`  tag and explicitly use the `resolute` portability image (we did something similar for the `oldest-boost` tag recently in https://github.com/Chaste/Chaste/pull/519).

PR #587, opened 2026-07-24
Speed up memtest ci by running continuous test pack across multiple runners

### LINKED ISSUE #584: Speed up memtest CI by running Continuous test pack across multiple runners
## Problem

The `memory-testing.yml` workflow (Valgrind memcheck on the `Continuous` test pack) currently takes **~3-4 hours per run**, almost entirely in the single `run memtest` step. From recent runs:

| Step | Duration |
|---|---|
| build continuous test pack | ~29 min |
| **run memtest** (`ctest -j4 -L Continuous` under Valgrind) | **~3h15m** |

This runs on an `ubuntu-22.04` GitHub-hosted runner (4 cores). The whole `Continuous` test pack runs through one `ctest -j4` invocation on a single runner.

## Proposal: split the memtest job by component

Every test is already labelled `<type>_<component>` e.g. `Continuous_cell_based`, so `ctest -L Continuous_cell_based` already works today with no CMake changes.

Turning `memory-testing.yml`'s single `build-and-test` job into a matrix, one leg per component (or a few balanced groups, e.g. `cell_based` alone, `heart` alone, everything else combined), each still on an identical 4-core ubuntu runner, running `ctest -L Continuous_<component>` instead of the whole `Continuous` label, should cut wall-clock time roughly in proportion to the number of runners.

fcooper8472:
See #584 

Split up the memory testing into `[cell_based, heart, crypt, lung, core]` and combine the results at the end. This should reduce runtime substantially.

Note that I've cherrypicked some commits from #581 as part of this work.

### DIFF
--- a/.github/workflows/memory-testing.yml
+++ b/.github/workflows/memory-testing.yml
@@ -9,6 +9,15 @@ on:
     branches:
       - develop
 
+env:
+  # Read by cmake/process_valgrind_output.py to label the generated report. head_ref and
+  # pull_request.head.sha are only set for pull_request events, so they fall back to the
+  # push/workflow_dispatch equivalents.
+  Chaste_GH_WORKFLOW_BRANCH: ${{ github.head_ref || github.ref_name }}
+  Chaste_GH_WORKFLOW_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+  # Core count of the runners used below, for cmake, make and ctest.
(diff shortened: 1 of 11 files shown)


jmpf (approved):
Unable to test this on 24.04.

- Code changes to shared pointers all look reasonable.
- Suppressions (2 for MPI and 1 for VTK) look minimal and similar to previous.

PR #590, opened 2026-07-28
Perf/avoid rtti

MILeach:


cell_based/src/cell/properties/mutations/ApcOneHitCellMutationState.cpp
```diff
@@ -36,7 +36,7 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "ApcOneHitCellMutationState.hpp"
 
 ApcOneHitCellMutationState::ApcOneHitCellMutationState()
-    : AbstractCellMutationState(3)
+    : CellProperty<ApcOneHitCellMutationState, AbstractCellMutationState>(3)
```
mirams, line 39: I don't really like this line, it might be unavoidable for some performance but needs a design discussion everyone has agreed on. Declaring a class to be a member of an abstract class that is templated on itself, and also needing to specify the abstract class that this would have been a member of, is a bit unusual for Chaste C++ anyway.

cell_based/src/population/MeshBasedCellPopulation.cpp
```diff
@@ -904,14 +904,38 @@ bool MeshBasedCellPopulation<ELEMENT_DIM,SPACE_DIM>::GetOffsetNewBoundaryNodes()
 
 
 
+template<unsigned ELEMENT_DIM, unsigned SPACE_DIM>
+typename MeshBasedCellPopulation<ELEMENT_DIM, SPACE_DIM>::MeshPeriodicityType
+MeshBasedCellPopulation<ELEMENT_DIM, SPACE_DIM>::GetMeshPeriodicityType() const
```
mirams, line 909: This Get() seems to be doing a Set() too, and I'm not sure how that's allowed when method is marked const.

PR #595, opened 2026-07-28
Build/faster linker

MILeach:
- Adds CMake infrastructure to use `mold`/`lld` if available by default, if not will fallback to `ld` for linking

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -15,8 +15,13 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    # Do not move this job to a plain Ubuntu runner with `apt install chaste-dependencies`.
+    # That was the original setup and was deliberately abandoned in #32: Ubuntu's system
+    # PETSc lacks ParMetis and Hypre, which stops PetscMatrixPartitioning from running and
+    # silently under-reports coverage rather than failing. The portability-* images build
+    # PETSc from source with --download-parmetis=1 --download-hypre=1.
     container:
-      image: chaste/runner:coverage-runner
+      image: chaste/runner:portability-tip
       env:
         RUNNER_OFF: 1
       options: --user 0 --cpus 4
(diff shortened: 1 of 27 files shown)


fcooper8472:
Update: since the earlier review of this PR, the following changes have been made on this branch.

1. Reverted the original CI workflow edits from this branch and merged `develop`'s CI overhaul in.
2. Replaced the always-on `-fuse-ld=` linker flag hack in `cmake/Modules/ChasteCompilerFlags.cmake` with a configurable `Chaste_FAST_LINKER` cache option: `AUTO`, `MOLD`, `LLD`, or `DEFAULT`.
   - `AUTO` (default): tries mold (>= 2.30.0, the version shipped with Ubuntu 24.04; older releases such as jammy's packaged 1.0.3 are known to produce binaries that crash at runtime), then lld, then falls back to the system default.
   - `MOLD` / `LLD`: force a specific linker; hard-errors at configure time if it isn't available, instead of silently falling back.
   - `DEFAULT`: always uses the system linker.
   - On CMake >= 3.29 with GNU, Clang, or AppleClang, the choice is applied via CMake's native `CMAKE_LINKER_TYPE` mechanism instead of hand-rolled `-fuse-ld=` flags. Older CMake, or IntelLLVM (undocumented for `LINKER_TYPE`), still uses the flags directly.
3. `tier-0-tests.yml` now forces a fast linker on each matrix leg (`gcc-debug` uses `MOLD`, `clang-release` uses `LLD`), so both paths are actually built and tested on every PR instead of only being reachable via `AUTO`.

**Benchmark results**

Local runs, minimum of repeated full `make clean` + rebuild cycles per linker:

| Configuration | Linker | Min build time | vs default |
|---|---|---|---|
| Continuous test pack, -j20 | MOLD | 193.2s | -30.2% |
| Continuous test pack, -j20 | LLD | 190.3s | -31.3% |
| Continuous test pack, -j20 | DEFAULT | 276.9s | (baseline) |
| heart + crypt + lung libraries, -j8 | MOLD | 313.6s | -13.5% |
| heart + crypt + lung libraries, -j8 | LLD | 313.6s | -13.5% |
| heart + crypt + lung libraries, -j8 | DEFAULT | 362.6s | (baseline) |

The isolated pure link time for `chaste_cell_based` (221 object files, 435MB) straight from already-compiled objects, swapping only the linker:

| Linker | Pure link time (no compilation) |
|---|---|
| bfd (default) | 24.5s |
| mold | 1.2s |
| lld | 1.5s |

Takeaways:
- Both fast linkers are a clear, consistent win over the system default linker on every configuration tested; the size of that win depends on how much of the given build is link-bound versus compile-bound (about 30% for the full Continuous pack, about 13.5% for the heart/crypt/lung libraries alone).
- Given the above, `AUTO` preferring mold when available, then lld, then the system default, seems sensible.

jmpf (approved):

Issue #486, opened 2026-01-20
Further parallelise cell-based chaste


MILeach:
## Goal

- Parallelisation of loops within chaste using OpenMP or standard library execution policies
	- [x] Choose representative simulation + 2017 cell-based comparison growing spheroid simulation. *Selected TestRepresentative3dNodeBasedSimulation*
	- [x] Profile to identify slow loops
	- [x] See what the compiler is doing already in an optimised build (e.g. are intrinsics used)
	- [x] CMake configuration for OpenMP within Chaste
	- [x] Search for opportunities where OpenMP could be applied
	- [x] Try out standard library execution policies + OpenMP
	- [x] Investigate OpenMP/CUDA/OpenCL flags for PETSc

## Good targets remaining
- AbstractCellProperty type checking, i.e. IsSame, IsType come up a lot and is doing typeid + strcmp
- GetCellUsingLocationIndex etc. very widely used

## PETSc GPU support
#528 

Status: PETSc built with GPU support, needs further investigation as there is no obvious performance change where I was expecting to see it

## OpenMP Acceleration
#527 

Status: 

## General Performance Improvements
#529 

Status: Various opportunities to avoid object copies in tight loops taken which help particularly on machines with lots of cores. Up to ~10-20% on machines where force calcs are now quick. Would be nice to speed up CellProperty type comparison (uses string comp)

PR #608, opened 2026-08-16
#486 NodeAttribute to make TestRepresentative3dNodeBasedSimulation quicker

stu-l:
This appears to make TestRepresentative3dNodeBasedSimulation run ~20% faster.  This is assuming duplicate neighbouring nodes aren't actually needed for something.

Note: node index is limited to a 0 to 4096 range, I have no idea if this is acceptable in the real world, seems to work in the tests (Continuous anyway).

### DIFF
--- a/mesh/src/common/NodeAttributes.hpp
+++ b/mesh/src/common/NodeAttributes.hpp
@@ -39,10 +39,141 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "UblasVectorInclude.hpp"
 #include "ChasteSerialization.hpp"
 #include <boost/serialization/vector.hpp>
+#include <boost/serialization/set.hpp>
+#include <bitset>
+#include <iterator>
 
 /**
  * A container for attributes associated with the Node class.
  */
+#define numNodes 4096
+
+template <std::size_t N>
+class MyBitset
+{
+public:
+  using value_type      = unsigned; 
+  using size_type       = std::size_t;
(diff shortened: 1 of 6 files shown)


stu-l:
Top five time wasters are reduced from this:
image
to this:
image

MILeach:
Hi Stu, 

Thanks for getting in touch with us & your work, sorry for not getting back to you sooner. We are just wrapping up a chunk of work on improving the performance of Chaste so it's good timing for you to find us - we'll take a look at this in context of the work we've done already and get back to you with some thoughts.

mesh/src/common/NodeAttributes.hpp
```diff
@@ -39,10 +39,141 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #include "ChasteSerialization.hpp"
 #include <boost/serialization/vector.hpp>
+#include <boost/serialization/set.hpp>
+#include <bitset>
+#include <iterator>
 
 /**
  * A container for attributes associated with the Node class.
  */
+#define numNodes 4096
```
fcooper8472, line 49: @stu-l thanks for putting this PR in. It's a neat approach in general, but unfortunately this hard-coded number of nodes is an absolute blocker for Chaste.

In cardiac simulations, for example, there could easily be many millions of nodes.

In general, it might not easily be calculable at compile time an upper bound on the number of nodes needed in the simulation.

Any solution we consider would have to work for an unbounded number of nodes, I think.

fcooper8472 (changes requested):

Issue #607, opened 2026-08-10
Coverage: uncovered code missed due to recent false green checks


fcooper8472:
See #599 for history.

The codecov project statuses on `develop` are failing: `core` 99.99%, `cell_based` 99.17%, `heart` 99.96% (`crypt` and `lung` are at 100%).

Data below is from the coverage artifact of the [Coverage run on `fa89f2b83`](https://github.com/Chaste/Chaste/actions/runs/30924353459).

## 1. A statement split over two lines reports as one line covered, one uncovered

Two sites, both a declarator on one line with its `=` initialiser on the next:

`heart/src/problem/HeartConfig.cpp:1133-1135`
```
1133          |     XSD_SEQUENCE_TYPE(cp::cell_heterogeneities_type::CellHeterogeneity)
1134        0 |     cell_heterogeneity
1135       18 |         = mpParameters->Simulation()->CellHeterogeneities()->CellHeterogeneity();
```

`heart/src/tissue/AbstractCardiacTissue.cpp:92-93`
```
  92        0 |     AbstractPurkinjeCellFactory<ELEMENT_DIM,SPACE_DIM>* p_purkinje_cell_factory
  93      595 |        = dynamic_cast<AbstractPurkinjeCellFactory<ELEMENT_DIM,SPACE_DIM>*>(pCellFactory);
```

gcc attributes "never-executed" to the declarator line. On a single line those blocks merge into one line record, which is hit. This is the line-level equivalent of the branch-coverage noise disabled in #605.

A scan of all component `src/` directories finds 101 declarator-then-continuation splits, ~35 of them this `T x` / `= expr;` shape, e.g. `mesh/src/vertex/MutableVertexMesh.cpp:2237,2239,2279,2281,3389,3392,3395`, `ode/src/solver/BackwardEulerIvpOdeSolver.cpp:71`, `continuum_mechanics/src/solver/IncompressibleNonlinearElasticitySolver.cpp:265,598`, `pde/src/postprocessing/Hdf5ToVtkConverter.cpp:162`. Only two report as uncovered today, but a compiler upgrade could turn more of them into failures.

## 2. An `if constexpr` branch that is instantiated but never executed

`heart/src/tissue/ExtendedBidomainTissue.cpp:61`: `rTo[0] = rFrom[0];` inside `if constexpr (SPACE_DIM == 1u)`. The 1-D instantiation compiles but is never called; the `else` branch on line 66 has 32 hits.

## 3. Genuine untested code

[...]

PR #621, opened 2026-09-01
#607 Join statements split over two lines and use auto to shorten them

fcooper8472:
Partially addresses #607

Joins 100 statements that were split over two lines at `=`, which gcov/lcov mis-attributes as one covered line and one uncovered line. This is the cause of the two live coverage failures in #607, plus ~35 latent ones of the same type. 2 further three-line statements fixed by hand.

While touching these lines, replaces the declared type with `auto`/`auto*`/`auto&`/`const auto&` where it's redundant or unspellable (casts, `new`, factory calls, iterators, `XSD_SEQUENCE_TYPE(...)&`), but leaves uBLAS `c_vector` and `c_matrix` alone.


mirams (approved):
Have we reported this as a bug up the tree to gcov/lcov team?

fcooper8472:
> Have we reported this as a bug up the tree to gcov/lcov team?

No, but we probably should!

PR #622, opened 2026-09-01
607 cover genuine gaps

fcooper8472:
More work towards #607


cell_based/src/cell_based_pde/ParabolicBoxDomainPdeModifier.cpp
```diff
@@ -317,7 +317,7 @@ Vec ParabolicBoxDomainPdeModifier<DIM>::InterpolateSolutionFromCellMovement(Abst
             // p_vtk_mesh_writer->WriteFilesUsingMesh(*p_deformed_mesh);
             // delete p_vtk_mesh_writer;
 
-            assert(0);
```
fcooper8472, line 320: @jmosborne this is the only source change for PDE code: this assert(0) will be compiled out in release mode, and the simulation will continue but be incorrect. So I've replaced it by terminate, so that if a user gets into this edge case, they will get a hard fail rather than a silent wrong solution

jmpf (approved):

cell_based/src/cell_based_pde/ParabolicBoxDomainPdeModifier.cpp
jmosborne: @fcooper8472 Very sensible Improvment

jmosborne (commented):
Cell Based PDE Tests are fine.

Issue #623, opened 2026-09-02
Coverage CI: lcov capture step is slow


fcooper8472:
The `Run coverage` step in `.github/workflows/coverage.yml` is slow:

- `lcov --capture` (geninfo) processes ~1048 `.gcda` files serially on one core; no `--parallel` is passed despite the container having 4 CPUs.
- lcov falls back to the pure-Perl `JSON::PP` module ("much slower than some alternatives") because no XS JSON module is installed in the `chaste/runner:portability-tip` image.

fcooper8472:
## Analysis

### Where the time goes
- **Serial**: no `--parallel` passed; `geninfo` defaults to 1 process despite the container's `--cpus 4`.
- **`JSON::PP`**: no `Cpanel::JSON::XS`/`JSON::XS` in the runner image, so gcov's JSON output is parsed in pure Perl (~1048 files).
- **Redundant second pass**: the separate `lcov --remove` re-parses and re-serialises the whole merged tracefile just to drop records.
- **Dead/dup patterns**: `/usr/*` is in both calls (`--remove` copy is a no-op); `Debug/*` and `Debug_*/*` match nothing with the Makefiles generator (SCons/`make coverage`-era cruft).

### On avoiding redundant processing
`geninfo --exclude` and `lcov --remove` are the same operation at the same point: both push onto `@exclude_file_patterns`, same glob->regex, matched after gcov has already run. Neither prevents the `gcov` fork + `.gcno` parse + JSON parse — `process_dafile` runs gcov unconditionally; excludes only drop records at merge time. So the ~1/3 of `.gcda` from test executables still get fully processed, and mostly must, because their kept header/template records are in the current report. The genuinely removable work is the second pass and the dead patterns, not the per-file gcov work.

### Recommendations (GCC + lcov, output byte-identical)
1. **Single pass**: fold all `--remove` patterns into `lcov --capture --exclude ...`, drop the second `lcov` call and the `/usr/*` dup and `Debug*` patterns. Same matcher, same match point -> identical output.
2. **`--parallel 4`** on the capture (explicit `4`, not `0`: geninfo's auto-detect reads host `/proc/cpuinfo`, ignoring the cgroup quota). Optionally also `parallel = 4` in `lcovrc` for the local `make coverage` target.
3. **Install `libcpanel-json-xs-perl`** in `chaste/runner:portability-tip` (lcov auto-selects it); optionally pin `json_module = Cpanel::JSON::XS` in `lcovrc`.
[...]

PR #624, opened 2026-09-02
#623 Speed up and unify lcov coverage capture

fcooper8472:
Addresses the configuration part of #623 

See https://github.com/Chaste/dependency-modules/issues/134 for the faster XML parser

### DIFF
--- a/cmake/Config/lcovrc
+++ b/cmake/Config/lcovrc
@@ -2,6 +2,20 @@
 # dedicated Chaste defaults for LCOV
 #
 
+# Paths excluded from coverage, in one place so both pipelines (CI codecov
+# .github/workflows/coverage.yml, local `make coverage` in root CMakeLists.txt) select the
+# same files. Applied at geninfo capture time, so `lcov --capture` needs no --exclude args
+# and no `lcov --remove` pass; also keeps system/module headers out of the merge, which is
+# what otherwise triggers the lcov 2.x `inconsistent`/`mismatch` errors (see below).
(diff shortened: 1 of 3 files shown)


fcooper8472:
The faster XML parser is now available.

Results:
Huge improvement in processing time:

34 minutes originally
16 minutes after running in parallel
1m21s after making the faster parser available

cmake/Config/lcovrc
```diff
@@ -2,6 +2,20 @@
...
+# and no `lcov --remove` pass; also keeps system/module headers out of the merge, which is
+# what otherwise triggers the lcov 2.x `inconsistent`/`mismatch` errors (see below).
+# NB: a command-line --exclude REPLACES this list, so neither pipeline passes one.
+exclude = /usr/*
+exclude = */modules/*
+exclude = */fortests/*
+exclude = */test/*
+exclude = */3rdparty/*
+exclude = */global/src/random/*
+exclude = */cxxtest/*
```
kwabenantim, line 17: Just checking that it's no longer relevant to exclude `'Debug/*' 'Debug_*/*'` as before

kwabenantim (approved):
Looks good to me. Just wondering if the 'Debug/' folders no longer need excluding.

cmake/Config/lcovrc
fcooper8472: Yes, that's deliberate. I don't think that will affect anyone, and definitely not CI. Thanks!

Issue #598, opened 2026-07-31
Audit cell_based test runtimes and cut them without losing efficacy


fcooper8472:
# Audit `cell_based` test runtimes and cut them without losing efficacy

## Problem

The `Continuous_cell_based` suite is slow, and it is disproportionately painful under memory testing. `ChasteMacros.cmake:184-187` wraps every test in

```
valgrind --tool=memcheck --track-fds=yes --leak-check=yes --num-callers=50
```

which typically costs 20-50x wall time, and `Chaste_MEMORY_TESTING_CPUS` defaults to 1. A suite that is a few minutes serially becomes hours under memcheck, so the memory-testing job is where runtime hurts most and where a reduction pays back the most.

## The time is concentrated

Measured from CTest cost data for a Debug build, 116 of the 119 `Continuous_cell_based` tests:

| | Serial time | Share |
|---|---|---|
| Whole suite | 413s | 100% |
| Top 5 tests | 91s | 22% |
| Top 10 | 144s | 35% |
| Top 20 | 218s | 53% |
| Top 30 | 266s | 64% |

Ten test binaries out of 119 account for over a third of the runtime, and twenty account for over half. Under memcheck those same proportions hold, so the audit does not need to touch most of the suite.

The slowest:

```
25.71  TestRepresentativeImmersedBoundarySimulations
22.75  TestOffLatticeSimulationWithNodeBasedCellPopulation
15.91  TestOffLatticeSimulationWithNodeBasedCellPopulationIn3d
13.56  TestRunningContactInhibitionSimulationsTutorial
13.09  TestOnLatticeSimulationWithCaBasedCellPopulation
11.73  TestOffLatticeSimulation
11.01  TestVolumeTrackingModifier
10.86  TestOffLatticeSimulationWithPdes
10.73  TestOffLatticeSimulationWithVertexBasedCellPopulation
 8.83  TestNodeBasedCellPopulation
 8.22  TestSimulationsWithEllipticBoxDomainPdeModifier
 7.73  TestCreatingAndUsingANewCellCycleModelTutorial
 7.52  TestVisualizingWithParaviewTutorial
 7.48  TestSimulationsWithParabolicBoxDomainPdeModifier
 7.45  TestRunningImmersedBoundarySimulationsTutorial
```

[...]

PR #627, opened 2026-09-02
Speed up cell based test suite

fcooper8472:
See #598 

Carefully speed up the longest-running cell_based tests. Many by shortening tutorial simulation times, with a clear note to the tutorial user that they can replace the simulation end time for a more useful output.

### DIFF
--- a/cell_based/test/simulation/TestOnLatticeSimulationWithCaBasedCellPopulation.hpp
+++ b/cell_based/test/simulation/TestOnLatticeSimulationWithCaBasedCellPopulation.hpp
@@ -135,43 +135,41 @@ public:
         // Create cell population
         CaBasedCellPopulation<2> cell_population(*p_mesh, cells, location_indices);
 
-        // Set up cell-based simulation
-        OnLatticeSimulation<2> simulator(cell_population);
-        std::string output_directory = "TestCaSingleCellRandomMovement";
-        simulator.SetOutputDirectory(output_directory);
-        simulator.SetDt(delta_t);
-        simulator.SetEndTime(delta_t);
-
-        // Add update rule
+        // Add update rule. Note that we call `CaBasedCellPopulation::UpdateCellLocations()` directly
+        // in the loop below, rather than driving this through an `OnLatticeSimulation`. Calling
+        // `OnLatticeSimulation::Solve()` in a loop num_runs times would re-run its per-call setup
+        // (creating and cleaning a new output directory and reopening writer files) num_runs times,
+        // which dominates runtime without adding anything this test checks; `UpdateCellLocations()`
+        // is the same method `Solve()` would call each timestep, and has no dependency on
+        // `SimulationTime` or the simulator. This does mean we skip `Solve()`'s per-timestep cell
+        // birth/death/population-update bookkeeping, but that is a no-op here (there are no cell
+        // killers and the cell is `DifferentiatedCellProliferativeType`, which never divides), and is
+        // covered thoroughly by the other tests in this file.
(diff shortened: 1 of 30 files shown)


cell_based/test/simulation/TestOnLatticeSimulationWithCaBasedCellPopulation.hpp
```diff
@@ -135,43 +135,41 @@ class TestOnLatticeSimulationWithCaBasedCellPopulation : public AbstractCellBase
...
+        cell_population.AddUpdateRule(p_diffusion_update_rule);
+
+        CellPtr p_cell = *(cell_population.Begin());
 
         for (unsigned i=1; i<=num_runs; i++)
         {
-            simulator.SetEndTime(delta_t*i);
-
-            // Run simulation
-            simulator.Solve();
```
fcooper8472, line 155: This test was calling `Solve()` 2000 times to verify the statistical properties of where cells move. But each single timestep was doing a whole load of setup, including writing `results_from_time_X` to disk. Instead, directly call `UpdateCellLocations`. Still testing the same statistical properties, but about 10x faster.

fcooper8472:
@jmpf and @mirams , keen to get your views on this approach.

Basically I think the cell based test suite is running WAY too long, with some individual tests essentially running full simulations that really oughtn't to be in a continuous or nightly suite.

I'm proposing, in most cases, that tutorials and representative simulations run in the test suite for a much shorter duration, and there's a really clear comment suggesting that users replace the end time with a more sensible value if they want to see the actual behaviour.

I've done this in quite a few places now, but won't go further without checking everyone's reasonably comfortable with this approach.

So far this has taken ~40 minutes off the cell_based memtest runtime (1h50 [previously](https://github.com/Chaste/Chaste/actions/runs/33643228598/job/100397969893), 1h10 [on this run](https://github.com/Chaste/Chaste/actions/runs/33680052143/job/100414155548)), and still significant scope for further improvements.

No drop in coverage - and that's a really good sign because I would hate to think that any tutorials (which don't actually test anything) were contributing to coverage results...)

mirams (approved):
We had all that stuff about making sure tests throw errors, and sorted cases where it wasn't doing (maybe parallel or something) so hopefully they all are now. Which should make TS_ASSERT_THROWS_NOTHING() redundant?

Probably not best practice to have all the `solve()` that will get copied and pasted wrapped in those?

mirams:
But I approve of the general approach of doing a couple of timesteps rather than running a load

fcooper8472:
> We had all that stuff about making sure tests throw errors, and sorted cases where it wasn't doing (maybe parallel or something) so hopefully they all are now. Which should make TS_ASSERT_THROWS_NOTHING() redundant?
> 
> Probably not best practice to have all the `solve()` that will get copied and pasted wrapped in those?

Yeah, I think I understand what you mean. Perhaps I get rid of all the `TS_ASSERT_THROWS_NOTHING`, but keep the general checks that we've actually reached the right simulation time? I think those hopefully give users some confidence that the simulation has actually run!

cell_based/test/simulation/TestOnLatticeSimulationWithCaBasedCellPopulation.hpp
jmpf: Nice to find and fix this.

jmpf:
I like that the simulations can be cut down... as long as they are still testing something reasonable -- like that mitotic divisions happened.
Meanwhile the TestOnLatticeSimulationWithCaBasedCellPopulation.hpp new approach of running the guts of the Solve() without doing all the checks is good.  I can't see any other reason to make so many multiple single-step calls to Solve().

PR #637, opened 2026-09-03
Run the np3 parallel test job on a GitHub-hosted runner

### LINKED ISSUE #636: Migrate the np3 parallel CI job from a self-hosted runner to a GitHub-hosted runner
## Background

`.github/workflows/ubuntu-np3.yml` runs the Continuous pack with `Chaste_NUM_CPUS_TEST=3` on `runs-on: self-hosted-np3`. It is the only workflow still on a self-hosted runner, which is a maintenance liability.

It is not already on a GitHub-hosted runner because the standard hosted Linux runner is a 4‑vCPU VM that is really **2 physical cores + SMT**. OpenMPI (both 4 and 5) sets `slots = detected cores = 2`, so `mpiexec -np 3` aborts with:

```
There are not enough slots available in the system to satisfy the 3
slots that were requested by the application
```

This is not a hard limit: there are 4 logical CPUs so 3 ranks should in principle fit.

## Proposed approach (minimal)

Pass an oversubscribe / hardware-thread pre-flag to `mpiexec`, scoped as tightly as possible so nothing changes for users on real machines.

Chaste already threads `MPIEXEC_PREFLAGS` into every parallel test invocation (`cmake/Modules/ChasteMacros.cmake:165`), and it is an un-set, user-overridable cache variable. So the smallest change is to set it in the np3 workflow's configure step and flip `runs-on`:

```yaml
runs-on: ubuntu-26.04

# ...
- name: cmake configure
  run: >
    cmake -DCMAKE_BUILD_TYPE=Release -DChaste_NUM_CPUS_TEST=3
    -DMPIEXEC_PREFLAGS=--use-hwthread-cpus ..
```

fcooper8472:
Closes #636

### DIFF
--- a/.github/workflows/ubuntu-np3.yml
+++ b/.github/workflows/ubuntu-np3.yml
@@ -15,7 +15,7 @@ jobs:
 
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci:off') }}
 
-    runs-on: self-hosted-np3
+    runs-on: ubuntu-26.04
 
     env:
       CHASTE_TEST_OUTPUT: ${{ github.workspace }}/chaste-test-dir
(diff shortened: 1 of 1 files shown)


fcooper8472:
Hangs when running np3. Not a viable solution, see https://github.com/Chaste/Chaste/issues/636

PR #658, opened 2026-09-06
#657 Introduce tiered, label-gated CI across all PR workflows

### LINKED ISSUE #657: CI: all workflows run in full, immediately, on every PR push
1. Every CI workflow triggers immediately, in full, on every PR push, with no distinction between cheap and expensive checks. 17 of the workflows in `.github/workflows/` declare an unscoped `pull_request` trigger (`branches: ["**"]` or equivalent): `ubuntu-tests.yml`, `ubuntu-2204-compilers.yml`, `ubuntu-2404-compilers.yml`, `ubuntu-basic-compatibility.yml`, `ubuntu-infrastructure.yml`, `ubuntu-np3.yml`, `coverage.yml`, `memory-testing.yml`, `profiling-gperftools.yml`, `profiling-gprof.yml`, `portability.yml`, `project-tests.yml`, `pychaste-ubuntu-tests.yml`, `doxygen.yml`, `update-tutorials.yml`, `update-how-tos.yml`, `update-pychaste-tutorials.yml`. Fast smoke-test-equivalent checks and slow deep-analysis checks (valgrind, coverage, profiling, portability) all fire concurrently on the same push.

2. `ubuntu-tests.yml` itself bundles a fast signal together with a substantially more expensive one: a 3-leg compiler/OS matrix (gcc/22.04, icx/22.04, gcc/24.04), each leg running three separate `ctest` label passes (`Continuous`, `Nightly`, `Parallel`). See `ubuntu-tests.yml:20-33` (matrix definition) and `:116-124` (three sequential ctest invocations per leg).

3. None of the PR-triggered workflows depend on one another — no `needs:` or `workflow_run` relationship exists between any of the 17 workflows in (1). An expensive workflow (e.g. `memory-testing.yml`'s valgrind pass, or `coverage.yml`) runs to completion even when a cheap check has already failed on the same commit.

4. The only existing opt-out mechanism, the `ci:off` PR label, is implemented as a near-identical `if:` guard copy-pasted into ~16 separate workflow files, and has already drifted once: `memory-testing.yml:25` and `:97` combine the guard with `!cancelled()`, while the guard in e.g. `ubuntu-tests.yml:16` and `coverage.yml:14` is a plain one-liner. There is no single place to read or change the label logic.

5. There is no opt-in mechanism. `ci:off` can only remove checks from a run; there's no way to start light and deliberately escalate to the full check set later. As a result, branch protection cannot require the expensive checks (coverage, memory-testing, profiling, portability) to pass before merge without also forcing them to run, unconditionally, on every push to every PR.

fcooper8472:
See #657.

> **Updated 2026-09-15** to reflect the final two-tier design merged in this PR. The changes described below were made in commit [e4cb2bbf2](https://github.com/Chaste/Chaste/commit/e4cb2bbf2b4560ef155d5933a47a1ddebff681a2), which replaced the three-tier (`ci:tier1`/`ci:tier2`) scheme originally proposed with a simpler two-tier one. The discussion below the fold still reflects that original proposal and how it evolved.

## What this does

Every PR-triggered workflow now runs through a shared gate (`_ci-gate.yml`) that decides, from the PR's labels, which jobs run. Work is split into two tiers:

| Tier | Covers | Opt-in |
|------|--------|--------|
| **T0** | full build and test (gcc debug + clang release), infrastructure and whitespace, doxygen coverage | always, unless `ci:off` |
| **T1** | everything else: latest-compiler sweep, a full build-and-test with the Intel compiler, np3 parallel, coverage, memory testing, portability, profiling (gperftools and gprof), project tests, pychaste tests, how-to / user-tutorial / pychaste-tutorial generation | label `ci:all` |

## Naming

Workflow files are prefixed `tier-N-`, workflow names are `TN: <Area>`, and every job name starts `TN · …`. The tier is visible in the file list, the Actions sidebar, and the PR checks, so there is no guessing whether a tier has run.

A gated job that runs shows as `TN · <name> (ran)`. When its tier is not enabled it shows once as a bare `TN · <name>` (skipped). Matrix jobs carry one short slug per leg rather than a dump of every matrix value:

- `T1 · compilers (clang-22-debug)`
- `T1 · portability (01-no-cvode-vtk)`
- `T1 · project tests (CellBasedComparison2017)`
- `T0 · build & test (gcc-debug)`

## Default behaviour

- By default only T0 runs; T1 jobs appear as skipped. T0 is now a full compile-and-test (not just a smoke check) on two compilers, so most changes get real signal without opting into anything.
- Requesting a review automatically adds `ci:all` if it isn't already present, so the full suite is guaranteed to have run before a human approves. This is the only place a developer needs to interact with the tiered-CI labels for the normal author → review → merge flow.
- `push` to `develop`, manual `workflow_dispatch`, and scheduled runs ignore labels and run everything.
- `ci:off` also disables T0, for docs-only changes or work in progress you do not want burning runners.

## Completed checks are not overwritten

Every gated job runs under a one-entry matrix. A later run that skips the job posts a separate bare `TN · <name>` row and cannot replace the `TN · <name> (ran)` result from the run that did the work. So:

- Adding `ci:all` does not re-run or cancel a T0 that's already running or finished.
- Toggling an unrelated label mid-run flips nothing.
- Because GitHub can't trigger a workflow only for the specific label that changed, every PR-triggered workflow still fires on any label event; the gate is what makes an irrelevant one a no-op skip rather than a real re-run.

## Cancelling in-progress runs

Each tiered job has a `concurrency` group keyed on the PR. A newer run supersedes the older one, so pushing twice in quick succession does not leave stale jobs running. Matrix legs are keyed individually so they do not cancel each other. A labeled or unlabeled event only cancels a tier it actually changes.

## Compilers

The compiler sweep now targets only the latest available versions on the newest supported Ubuntu LTS (gcc and clang, debug and release), rather than a spread of historical versions across multiple OS releases. The one full build-and-test using the Intel compiler lives in its own workflow (`tier-1-intel.yml`), separate from the compile-only sweep, so a build check is never confused with a real test run.

## Publishing split out

Doxygen and tutorial publishing used to run inside the tier-1 workflows behind `github.event_name == 'push'` guards, leaving skipped "publish …" rows on every PR. That work now lives in its own workflows:

- `publish-on-develop.yml` (`push` to `develop`, plus manual `workflow_dispatch`): publishes doxygen-latest and regenerates the how-to and tutorial pages onto the website.
- `doxygen-release.yml` (`workflow_dispatch` with a name input): builds the release documentation and uploads it as an artifact.

The corresponding tier workflows are now PR-only checks: they verify the generators run, they no longer publish. The doxygen coverage gate for `develop` moves into `publish-on-develop.yml`.

## Also here

- `macos.yml`, `vtk-version-monitoring.yml` and `ubuntu-basic-compatibility.yml` are removed: served their purpose or low value.
- The redundant `branches: ['**']` filter is dropped from every `pull_request` trigger: it is a no-op.
- Job names are audited to say what each job actually does.

---

<details>
<summary>Original proposal (three-tier <code>ci:tier1</code>/<code>ci:tier2</code> scheme, superseded by the update above)</summary>

Every PR-triggered workflow now runs through a shared gate (`_ci-gate.yml`) that decides, from the PR's labels, which jobs run. Work is split into three tiers:

| Tier | Covers | Opt-in |
|------|--------|--------|
| **T0** | build and test, all packs | always, unless `ci:off` |
| **T1** | infrastructure and whitespace, doxygen coverage, np3 parallel, u2204 and u2404 compilers, how-to / user-tutorial / pychaste-tutorial generation | label `ci:tier1` (or `ci:tier2`) |
| **T2** | coverage, memory testing, portability, profiling (gperftools and gprof), project tests, pychaste tests | label `ci:tier2` (implies T1) |

This scheme, and the discussion below, drove the design but was replaced before merge by the simpler two-tier version described above — kept here for the record.

</details>


fcooper8472:
@kwabenantim @jmpf @mirams I've tagged you to review as you're likely to have opinions on this. This is the biggest overhaul to CI since we switched to GitHub.

I want to significantly reduce the amount of time spent running checks. My philosophy is that - by default - we just "build and test", and none of the bells and whistles need to run at all until we're reasonably happy that the change is ready to ship. Only when we think we're finished do we really need to check portability, do memory testing, coverage etc.

I'm proposing this for two reasons:

1. To reduce the number of minutes we're using, for environmental reasons
2. When we open many PRs e.g. Claude being asked to fix 6 different little things, we hit the max concurrency limit, and checks wait in a pool for ages

To highlight the environmental side of things:

| Window                          | Runner-hours | Approx CO₂e        |
|---------------------------------|--------------|--------------------|
| Last 7 days (Sep 1–7)           | ~1,436 h     | ~55 kg  |
| Last 30 days (Aug 8 – Sep 7)    | ~2,950 h     | ~120 kg |
| Year to date (Jan 1 – Sep 7)    | ~10,644 h    | ~425 kg |


### Tiering

This PR tiers the checks T0, T1, T2, and by default only T0 runs. You opt in to T1 and T2 by adding the labels `ci:tier1` and `ci:tier2`, or you opt out of all checks, including T0, by adding `ci:off`. You can, of course, select either `ci:off` or `ci:tier2` when you **create** the PR, if you really want none of, or all of, the checks from the start.


### Cancel in-progress

If you push twice in quick succession, your second push will now kill the in-progress checks from the first push. A no-brainer (thanks @kwabenantim for suggesting).

mirams:
Can we be more conventional/seamless with github by only running T0 when the github 'Draft' setting is on a PR, and triggering the full T0->T1->T2 when 'Draft' label is removed? To avoid people having to learn/remember these new tags?

Basically, I would think you still want to 'waterfall through T0, T1, T2' automatically, without developer needing to know that's happening, but only do T1 if T0 passes etc. (maybe even simpler first tier can just be "does `global` compile and test OK on default ubuntu" and fast infrastructure script checks etc.). 

I had actually assumed we were cancelling tests for superceded PRs anyway, that makes sense

mirams:
(it has been nice that devs have not had to look at failing memory and coverage tests and work out which PR caused them, wouldn't want to have to do that again if passing Tier0 things get merged in to develop, so think we want some way to ensure that the full test suite runs on each PR before it can be merged in)

fcooper8472:
> Can we be more conventional/seamless with github by only running T0 when the github 'Draft' setting is on a PR, and triggering the full T0->T1->T2 when 'Draft' label is removed? To avoid people having to learn/remember these new tags?

I personally don't like this option. I think creating a PR as a draft is more hidden than the labels. This wouldn't have prevented Friday's chaos of Calude opening 6 or so PRs and everything running, because none of those were opened as drafts. I think it's important that the default becomes "not everything runs", or we're not much better off than before.

You either have to remember to create as draft, OR you have to remember to add a label.

fcooper8472:
> Basically, I would think you still want to 'waterfall through T0, T1, T2' automatically, without developer needing to know that's happening

I think I've come around to disagreeing with that. This does mean that, assuming no failures, you're running all the really heavy workflows on each commit that's pushed to a PR. I like the aspect of this design that allows you to opt out of running the heavy stuff until the feature is nearly ready!

mirams:
Is it possible to put some hook in then when you press "merge" that checks for presence of a `ci:<next tier>` label, and if it isn't there adds it and runs the next level of tests?

fcooper8472:
> (it has been nice that devs have not had to look at failing memory and coverage tests and work out which PR caused them, wouldn't want to have to do that again if passing Tier0 things get merged in to develop, so think we want some way to ensure that the full test suite runs on each PR before it can be merged in)

A review will still be required before merging, so the only additional "risk" here is that one of us reviews without checking what's run.

But now it's easier than before, in that if the summary says that anything is skipped, the reviewer shouldn't approve. Previously we expected some "skipped" checks (e.g. not uploading how-tos unless it's a push to develop), but that's refactored now. Any "skipped" checks now means you've not run everything.

mirams:
Simplest then is a failing test on Tier 0 that when you click on it says "Now add the Tier 1 label" ?

I am basically railing against any additional "Oh but to become a Chaste developer you also need to know to do this non-standard thing..." ;-)

mirams:
Or require ci:tier2 before merge is available: https://stackoverflow.com/questions/69697522/dont-enable-the-green-merge-button-unless-the-pr-has-ready-for-merge-label (has same problem that someone new won't know why the merge isn't green)

jmpf:
> Or require ci:tier2 before merge is available: https://stackoverflow.com/questions/69697522/dont-enable-the-green-merge-button-unless-the-pr-has-ready-for-merge-label (has same problem that someone new won't know why the merge isn't green)

So long as Claude/us don't start automatically adding the tag at the beginning of the process, this would be a nice compromise.  You can still merge something with bad coverage/memory, but you have to know that it has been tested or is being tested...

fcooper8472:
> Or require ci:tier2 before merge is available: https://stackoverflow.com/questions/69697522/dont-enable-the-green-merge-button-unless-the-pr-has-ready-for-merge-label (has same problem that someone new won't know why the merge isn't green)

Unfortunately we can't do this cleanly. I tried this already, and it requires you to have a failing test before it's ready to merge. Then every single PR with development not yet complete will be a red cross.

fcooper8472:
I've just added a load of Tier 2 checks to the branch protection rules - it should now be impossible to merge if the Tier 2 checks haven't run and passed.

fcooper8472:
> I've just added a load of Tier 2 checks to the branch protection rules - it should now be impossible to merge if the Tier 2 checks haven't run and passed.

This isn't good, either: it'll just show a bunch of required tests pending - potentially for weeks or months..!

image

fcooper8472:
I think we're essentially in one of two places:

1. A system where everything runs (eventually) on every commit that's pushed (even if via a waterfall); or
2. Reviewers have to check that the required workflows have run

I'm personally leaning heavily towards option 2. In any case, if we actually let something through because we approved something without `ci:tier2` we will still get feedback an hour later when the workflows fail on push to develop. So we will quickly catch any mistakes that do crop up.

mirams:
What's wrong with the idea of one failing test on a lower tier test pack (perhaps only if nothing else fails to avoid annoyance if that's possible?) that is just checking for presence of higher tier label? So things stay unmergeable until all tests are run and it gives clear instructions on what to manually do to enable the next level of tests?

fcooper8472:
> What's wrong with the idea of one failing test on a lower tier test pack (perhaps only if nothing else fails to avoid annoyance if that's possible?) that is just checking for presence of higher tier label?

I think it would really annoy me to be working on a feature for potentially a long time and having a constant red cross next to the workflow. Might just be me - but I think that would nag at me big time. I think I'd rather have "green = passing all the workflows I've currently enabled" and "red = I need to fix something" rather than "red = the feature isn't ready yet to turn on the final tier".

Happy to consider that, though, if other people have strong opinions!

kwabenantim:
> This isn't good, either: it'll just show a bunch of required tests pending - potentially for weeks or months..!

> What's wrong with the idea of one failing test on a lower tier test pack (perhaps only if nothing else fails to avoid annoyance if that's possible?) that is just checking for presence of higher tier label? So things stay unmergeable until all tests are run and it gives clear instructions on what to manually do to enable the next level of tests?

@fcooper8472 what are your thoughts on adding a single `T2` test that is required to merge which only checks that all the labels have run? That way there's just one thing that is perpetually pending - like a constant reminder to run everything before merging.

mirams (approved):
I'll approve so you can merge and see what happens!

mirams:
Latest idea, use the "invite review" webhook to add the label to run all tests:
```yaml
name: Label on review request

on:
  pull_request:
    types: [review_requested]

permissions:
  pull-requests: write

jobs:
  label:
    runs-on: ubuntu-latest
    steps:
      - name: Add ci:tier1 label
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: gh pr edit "$PR" --add-label "ci:tier1" --repo "$GITHUB_REPOSITORY"
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          PR: ${{ github.event.pull_request.number }}
```

fcooper8472:
> Latest idea, use the "invite review" webhook to add the label to run all tests:

That's a fantastic idea! Great thinking. I'll make that change tomorrow and then merge the PR and we can test it out for a little while.

mirams:
I'm thinking the basic flow should now be "waterfall" (possibly concurrent tests in quite a few more tiers, which stops incrementing tiers after any failure). But for pre-review we stop at tier 2 below maybe (regardless of results), and then review request adds a  `ci:full_waterfall`.

So I'd maybe have

- Tier 0: everything compiles and global runs in latest ubuntu, any cheap infrastructure tests that take less than a minute
- Tier 1: Continuous tests, all infrastructure, doxygen coverage all in latest ubuntu
- Tier 2: np3 parallel, Nightly, 

(above all waterfall through on every commit if it all passes, below only after the `ci:full_waterfall` label is added)

- Tier 3: all ubuntu variants, portability
- Tier 4: profiling, weekly tests
- Tier 5: coverage
- Tier 6: memory
- Tier 7: project tests
- Tier 8: PyChaste tests

or something like that. Could rejig the order of all these on "most common/cheap failures first" to minimise total computation and carbon emissions of a PR! (could even dig into the logs to do this as an optimisation problem!)

mirams:
...I'll ask Claude to have a go at analysing run times and failure rates....

fcooper8472:
> I'm thinking the basic flow should now be "waterfall" (possibly concurrent tests in quite a few more tiers, which stops incrementing tiers after any failure). But for pre-review we stop at tier:N (regardless of results), and then review request adds a `ci:full_waterfall`.
> 
> So I'd maybe have
> 
> * Tier 0: everything compiles and global runs in latest ubuntu, any cheap infrastructure tests that take less than a minute
> * Tier 1: Continuous tests, all infrastructure, doxygen coverage all in latest ubuntu
> * Tier 2: np3 parallel, Nightly,
> 
> (above all waterfall through on every commit if it all passes, below only after the `ci:full_waterfall` label is added)
> 
> * Tier 3: all ubuntu variants, portability
> * Tier 4: profiling, weekly tests
> * Tier 5: coverage
> * Tier 6: memory
> * Tier 7: project tests
> * Tier 8: PyChaste tests
> 
> or something like that. Could rejig the order of all these on "most common/cheap failures first" to minimise total computation and carbon emissions of a PR! (could even dig into the logs to do this as an optimisation problem!)

@mirams this is too complicated!! And the whole point of this was to reduce the amount of CI that's running unnecessarily. A full waterfall on every pushed commit is - I feel quite strongly - unnecessary. Let's please discuss this before you spend a lot of effort with claude on this. I have invested quite a bit of time and there are various limitations and pitfalls that I've already spent time falling down on this.

mirams:
It's even less running than you've proposed isn't it, splitting down the T0 into more tiers (so less runs unnecessarily if simplest tests fail), and the same for the T1-T2 half of waterfall checks which only happen when label(s) added (so again, less runs unnecessarily). Having a more linear flow with less running concurrently will mean a typo update a few minutes later cancels previous waterfall (and again less runs unnecessarily). 

We would still want the full suite running on an almost-ready-to-merge commit to a PR wouldn't we?

mirams:
My latest thoughts after asking Claude to analyse runtimes and failure rates... (deleted comment above)

I would prefer just two tiers rather than 3, with the "full suite" label added automatically on review request, if developer hasn't done it before. That way there's no new Chaste-specific thing that a new developer has to learn, we really need to guard against more and more barriers to entry.

T0 could just be Infrastructure tests (cheap and fail fairly often for copyright, doxygen(?) etc.) and one basic ubuntu-tests.
T_full = everything

Things to ponder:
 * Should we cancel the rest of the tests when one fails? I think people probably do a new commit and push with every fix, rather than trying to do all at once, so this could save a lot of compute (38%) and Claude thinks the median time to get to a green tick or red cross would then go down from 243 → 32 min, i.e. when there's a failure at the moment it's detected fast-ish but we still do full memory tests etc. So that could save compute, but potentially a handful of commits rather than just one to get 'clean' again.
 * The basic Ubuntu tests currently run three variants (gcc/icx with noble/jammy). Claude thinks this is a bit of a waste of resources as a couple have never disagreed (over last 300+ commits), and the icx picks up a couple more clang tests or something, so 2/3 of the ubuntu variants could wait for full portability builds. Claude thinks this would pick up 2/3rds of failures with this lot in the table below getting through. 

Goes red | Pushes | % of failures | % of commits | Only red workflow | % of failures | % of commits
-- | -- | -- | -- | -- | -- | --
PyChaste | 40 | 21% | 12% | 22 | 12% | 7%
Memory | 27 | 14% | 8% | 14 | 7% | 4%
Projects | 24 | 13% | 7% | 15 | 8% | 5%
Portability | 22 | 12% | 7% | 14 | 7% | 4%
Coverage | 14 | 7% | 4% | 8 | 4% | 2%
Tests, other legs | 8 | 4% | 2% | 4 | 2% | 1%
gprof / np3 / compilers | 10 | 5% | 3% | 3 | 2% | 1%
All that get through | 104 | 55% | 32% |   |   |  

I guess a lot of this could be non-PR-related PyChaste failures, so pinch of salt etc... I expect it would pick up nearer 80% of problems with just this simple T0.

fcooper8472:
@mirams all sounds sensible to me! I've done pretty much what you suggested.

First, collapsed the scheme from three tiers to two, and:

- T0 now runs the whole build + test on gcc-debug and clang-release. I thought we should at least do one debug and one release build, and that gave an opportunity to get two compilers in to T0 as well.
- infra/whitespace and doxygen-coverage checks also now in T0.
- T1 is now "everything else", one label (ci:all)
- Compilers simplified: we now just test the latest GCC and Clang, in 26.04, so much less compiling than before, and now in T1.
- Intel is now a single compile/test in T1
- New label-on-review-request.yml: auto-adds ci:all when a reviewer is added, and the full suite is guaranteed to run before a human approves.

mirams (approved):
Great stuff!

Issue #548, opened 2026-06-16
File/directory creation doesn't set explicit permissions


mirams:
See also #341 

Using the VSCode remote method. About 12 tests fail in total, some in subprocess exceptions. I got bored of running loads, so just did `Continuous_global` for the below. Tried with Debug and Release builds, didn't make any difference. Tried every combo of these, basically the same problems:

* Docker VMM 
* Apple Virtualisation w Rosetta VirtioFS 
* Apple Virtualisation w/o Rosetta VirtioFS 
* Apple Virtualisation w Rosetta gRPC FUSE 
* Apple Virtualisation w/o Rosetta gRPC FUSE 
* Apple Virtualisation w Rosetta osxfs (Legacy) 
* Apple Virtualisation w/o Rosetta osxfs (Legacy)

These all give some failures in `global`: 
 * TestArchivingHelperClasses (always)
 * TestLogFile (just Docker VMM*)
 * TestFileFinder (sometimes)
 * TestOutputFileHandler (always)

Give these sorts of errors:
```
***** TestArchivingHelperClasses.hpp *****


Entering TestArchiveLocationInfoMethods
Passed
Entering TestArchiveLocationInfoProcessUniqueNaming
Passed
Entering TestProcessSpecificArchive
Passed
Entering TestArchiveOpenerReadAndWrite

/home/chaste/src/global/test/TestArchivingHelperClasses.hpp:158: Error: Test failed: 
Chaste error: ./global/src/checkpointing/ArchiveOpener.cpp:208: Failed to open main archive file for writing: /home/chaste/output/archiving_helpers/archive_opener.arch
Failed
Entering TestBinaryArchiveOpenerReadAndWrite

/home/chaste/src/global/test/TestArchivingHelperClasses.hpp:210: Error: Test failed: 
Chaste error: ./global/src/checkpointing/ArchiveOpener.cpp:208: Failed to open main archive file for writing: /home/chaste/output/archiving_helpers/archive_opener_binary.arch
Failed
Entering TestArchiveOpenerExceptions

/home/chaste/src/global/test/TestArchivingHelperClasses.hpp:264: Error: Test failed: filesystem error: status: Permission denied [/home/chaste/output/archiving_helpers/archive_opener.arch.0]
Failed
Entering TestSpecifyingSecondaryArchive
Passed
Entering TestOpenFutureBoostArchive
Passed

Failed 3 of 8 tests
Success rate: 62%
```
and
```
[...]

PR #660, opened 2026-09-16
Comment out output folder bind mount to avoid permission problems and test failures on macOS

bdevans:
This comments out the config lines that previously bind mounted the devcontainer's output folder onto the host. While convenient, this causes some tests to fail due to problems with the implementation of file permissions on macOS.

See the following issues for more information:
https://github.com/Chaste/chaste-docker/issues/25 https://github.com/Chaste/Chaste/issues/548


fcooper8472 (approved):

mirams:
I'm a bit confused by the output of the new workflow @fcooper8472 - something seemed to fail when this was merged in, but in the above it all looked 'green' if you click on 'Show Details'. But then it is showing a lot of things skipped and then repeated but not skipped. So not sure what's going on.

fcooper8472:
The T0 are skipped deliberately: they already ran as default, so they don't need to re-run when the label is applied. Annoyingly (and this is a genuine limitation) the UI only shows the most recent, which in this case is "skipped". Where did it fail?

fcooper8472:
Ah the failing one on develop is a memory leak. V strange because it passed on the PR (with identical code): https://chaste.github.io/memory-testing/

So not caused by the new workflows, but mysterious nonetheless. Let me just try re-running and see if it resolves? If it does let's keep an eye out for any more sporadic failures...

mirams:
> The T0 are skipped deliberately: they already ran as default, so they don't need to re-run when the label is applied. Annoyingly (and this is a genuine limitation) the UI only shows the most recent, which in this case is "skipped". Where did it fail?

Yeah, but it is also showing things marked T1 as skipped and then again as not skipped and green... weird to me that it wouldn't overwrite these (if skipped pre-label) when they are later run with the label. Makes it a bit messy to read.

Is it because it add "(ran)" to the name (was that you or Github?). Hmmm, no that's only the bottom ones. It's showing 105 checks 

image

When you click 'View Details' above it looks as if lots of things didn't run, but then their constituent parts did. Bit confusing.
image

fcooper8472:
@mirams , from Claude:

> I checked the actual check data for this PR — it's not the same check showing two results. Every gated job posts two distinctly-named rows: a bare-name one (`T1 · coverage build & test`, skipped) from before the label was added, and a suffixed one (`T1 · coverage build & test (ran)`, or a per-leg name, succeeded) from after. No check name has two conclusions; the "105 checks passed" total is accurate. The confusion is just that the two rows for a given job aren't adjacent in the flat list, so scrolling past a run of bare-skip rows before reaching their `(ran)` counterparts looks like a wall of "didn't run."
> 
> Why it's built this way: GitHub can't trigger a workflow only for the specific label that changed, so any label add/remove re-fires every PR workflow, including ones a label is irrelevant to. Early on we had the job re-run for real on every such event, wasting exactly the compute this scheme exists to cut. The fix was to key every gated job's check name on whether it actually ran (bare name = skip, `(ran)`/per-leg name = real run), so a skip from an irrelevant re-fire can never overwrite a real result, in either direction. The cost is the extra bare row sitting alongside it — cosmetic, not a correctness issue.

mirams:
Thanks Claude!

mirams:
I guess if there's a new commit with the label added it will be more common sense and just show everything ran

PR #663, opened 2026-09-16
Add weekly zizmor audit dependabot for GitHub actions workflows

### LINKED ISSUE #661: Add weekly zizmor audit + Dependabot for GitHub Actions workflows
## Summary

Chaste.github.io recently added workflow-hardening tooling (Chaste/Chaste.github.io#218, #224, #227) that would be worth replicating here: a weekly [zizmor](https://github.com/zizmorcore/zizmor) audit of our GitHub Actions workflows, plus Dependabot to keep pinned action versions current. This repo is public, so GitHub code scanning (Advanced Security) is free here, same as it was once Chaste.github.io went public.

### 1. Weekly zizmor workflow
- Trigger: `schedule` (weekly) + `workflow_dispatch`. No `pull_request`/`push`
- Steps: run zizmor and upload security checks directly to GitHub (always passes with a green tick, even if problems), then
- run it again to see failures, uploading annotations to the failing job

### 2. Dependabot for github-actions
- `.github/dependabot.yml`, `package-ecosystem: github-actions`, weekly schedule.
- `cooldown.default-days: 14`: delays routine (non-security) version updates by 2 weeks, giving time for a compromised release to be caught before we adopt it.
- `groups.actions.patterns: ["*"]`: bundles multiple action bumps into a single PR per run instead of one PR per action.
- Freely allow major-version bumps (no `ignore` rules).

fcooper8472:
Closes #661

Summary of changes:

1. New weekly zizmor audit workflow (.github/workflows/zizmor-audit.yml): runs zizmor against .github/workflows on a weekly schedule and via workflow_dispatch, uploads SARIF results to GitHub code scanning, and separately runs zizmor in the github annotation format so findings show up as inline annotations and fail the job. No auto-fix in CI: fixes are reviewed by a human before merging, not run unattended.

2. New Dependabot config (.github/dependabot.yml): weekly updates for the github-actions ecosystem, grouped into a single PR, with a 14 day cooldown on routine version updates so a newly published release has time to be caught if it turns out to be compromised. Security advisory updates are not affected by the cooldown.

3. Fixed all findings from the first zizmor run (started at 222, now 0):
   - artipacked (credential persistence): added `persist-credentials: false` to checkout steps across the workflows. For the 5 steps that check out a second repo with an explicit PAT and push back to it later in the same job (doxygen-latest and website-docs in publish-on-develop.yml, and the results-publishing jobs in tier-1-memory-testing.yml, tier-1-profiling-gperftools.yml and tier-1-profiling-gprof.yml), the token is instead re-injected via git remote set-url immediately before the push, so it is not left sitting in git config for the rest of the job.
   - unpinned-uses: pinned all action references to commit SHAs, keeping the version tag as a trailing comment for readability.
   - self-repository: switched the local reusable workflow reference in tier-1-update-tutorials.yml from ./... to the dedicated $/... syntax.
   - excessive-permissions: added an explicit "permissions: contents: read" block to the 17 workflow files that had no permissions declared. Checked each one first: none of them use anything that needs more than read access (artifact upload/download and codecov do not consume GITHUB_TOKEN scopes, and the cross-repo pushes authenticate with their own PATs, not GITHUB_TOKEN).


.github/workflows/zizmor-audit.yml
```diff
@@ -0,0 +1,42 @@
...
+          persist-credentials: false
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+        with:
+          enable-cache: false
+
+      - name: Generate SARIF for code scanning
+        run: |
+          uvx zizmor --format sarif .github/workflows > results.sarif
```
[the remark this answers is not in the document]
fcooper8472: Fixed by changing instead to the zizmor action (pinned SHA) rather than invoking directly with uvx

.github/workflows/tier-0-doxygen.yml
```diff
@@ -43,7 +48,7 @@ jobs:
 
       - name: build doxygen
         run: |
-          ( cat Doxyfile ; echo "PROJECT_NUMBER=PR_branch::${{ github.head_ref }}" ) | doxygen-1.9.8/bin/doxygen - 2>doxygen-error.log 1>doxygen-output.log
+          ( cat Doxyfile ; echo "PROJECT_NUMBER=PR_branch::${GITHUB_HEAD_REF}" ) | doxygen-1.9.8/bin/doxygen - 2>doxygen-error.log 1>doxygen-output.log
```
bdevans, line 51: Doxygen still has a hard-coded version number here (and in some other places) - is that ok?

.github/workflows/tier-1-profiling-gprof.yml
```diff
@@ -130,26 +134,29 @@ jobs:
...
           token: ${{ secrets.PROFILING_GPROF_ACCESS }}
+          persist-credentials: false
 
       - name: copy results
         run: |
-          mkdir ${GPROF_DIR}/log-files/${{ needs.build-and-test.outputs.datetime }}
-          cp ./profile/* ${GPROF_DIR}/log-files/${{ needs.build-and-test.outputs.datetime }}/
+          mkdir ${GPROF_DIR}/log-files/${NEEDS_BUILD_AND_TEST_OUTPUTS_DATETIME}
+          cp ./profile/* ${GPROF_DIR}/log-files/${NEEDS_BUILD_AND_TEST_OUTPUTS_DATETIME}/
+        env:
```
bdevans, line 155: Should declaring the ENV variable come before it's used a few lines above?

.github/workflows/tier-1-profiling-gprof.yml
```diff
@@ -166,9 +173,14 @@ jobs:
...
+          git remote set-url origin "https://x-access-token:${TOKEN}@github.com/Chaste/profiling-gprof.git"
           git push
         working-directory: ${{ env.GPROF_DIR }}
+        env:
+          TOKEN: ${{ secrets.PROFILING_GPROF_ACCESS }}
 
       - name: results URL on chaste.github.io
         run: |
-          echo https://chaste.github.io/profiling-gprof/2025-09-29_15-55-53/${{ needs.build-and-test.outputs.datetime }}/index.html
+          echo https://chaste.github.io/profiling-gprof/2025-09-29_15-55-53/${NEEDS_BUILD_AND_TEST_OUTPUTS_DATETIME}/index.html
```
bdevans: Wot's this?! "2025-09-29_15-55-53"

bdevans (dismissed):
There are a few minor comments for details that are probably fine but that I wanted to flag. Otherwise, LGTM!

.github/workflows/tier-0-doxygen.yml
fcooper8472: I think we're safe because we're doing `wget https://github.com/doxygen/doxygen/releases/download/Release_1_9_8/doxygen-1.9.8.linux.bin.tar.gz`, and hopefully(?) that's unlikely to be replaced by something nefarious. But I guess it's impossible to be completely safe...

.github/workflows/tier-1-profiling-gprof.yml
fcooper8472: No, the env block is available throughout the step, and is calculated ahead of time.

.github/workflows/tier-1-profiling-gprof.yml
fcooper8472: uuh that looks suspicious!!

fcooper8472: Good catch - completely unrelated, and only means the url to click on has been incorrect.

fcooper8472:
```suggestion
          echo https://chaste.github.io/profiling-gprof/${NEEDS_BUILD_AND_TEST_OUTPUTS_DATETIME}/index.html
```

.github/workflows/tier-0-doxygen.yml
bdevans: Is it worth making it a variable for ease of updating in the future?

.github/workflows/tier-1-profiling-gprof.yml
bdevans: Fair enough, although the config would still look neater and more straight-forward if the ENV came before its use...

jmpf (approved):

PR #667, opened 2026-09-17
require newer mold on arm64 to avoid known thunk-generation crash (#666)

### LINKED ISSUE #666: mold crashes linking large libraries on ARM64 (upstream bug)
Summary: mold can crash linking large Chaste libraries on ARM64 (upstream bug).

What happened: after the fast-linker work landed, the arm64 leg of the develop Docker build failed. mold 2.40.4 crashed linking libchaste_mesh.so with an internal assertion in its range-extension-thunk code.

Root cause: known upstream mold bug (rui314/mold#1538), same class of issue as rui314/mold#1224. Fixed upstream by mold commit 5beb02714c, first released in mold 2.42.0.

Fix applied: cmake/Modules/ChasteCompilerFlags.cmake now requires mold >= 2.42.0 on aarch64/arm64 (still >= 2.30.0 elsewhere). AUTO falls back to lld automatically when the installed mold is too old, so this is not a hard failure, just a missed speedup until a newer mold is available.

Follow-up: chaste-docker's base image currently installs mold 2.40.4. Bumping it to >= 2.42.0 would let ARM64 Docker builds use mold again, but this is not urgent since lld is already used as a safe fallback.

References:
- https://github.com/rui314/mold/issues/1538
- https://github.com/rui314/mold/commit/5beb02714c7b7d6ddbe949772f4c9331ada41a43

fcooper8472:
See #666


bdevans (approved):
LGTM!

PR #669, opened 2026-09-17
Pin doxygen download in actions and check its checksum

### LINKED ISSUE #668: doxygen download in CI had no integrity check
Summary: doxygen download in CI had no integrity check.

What happened: three workflows (`doxygen-release.yml`, `publish-on-develop.yml`, `tier-0-doxygen.yml`) download a pinned doxygen release tarball via wget and run it, with no checksum verification. This predates the zizmor hardening in #663, which only covers `uses:` action references, not `wget` calls in run: steps, so this gap was not caught. @bdevans suggested this change while reviewing that PR.

Fix: add a sha256sum check after the download, and move the version and checksum into a job-level env block per workflow so they are easier to update together.

Note: the doxygen version is pinned intentionally, not tracked by dependabot, to keep generated HTML byte-stable for the github pages publish step. We should bump the version and checksum together only when a change is actually needed.

fcooper8472:
Closes #668


bdevans (dismissed):
This looks good but is there a way to abstract the version and checksum (along with other variables perhaps) into a repo-wide `.env` file (or equivalent) that multiple workflow files can point to?

fcooper8472:
> This looks good but is there a way to abstract the version and checksum (along with other variables perhaps) into a repo-wide `.env` file (or equivalent) that multiple workflow files can point to?

@bdevans yes, it turns out you can do that! How's the new version?

bdevans (approved):
Lovely stuff!

PR #676, opened 2026-09-21
Remove optional Xerces, CVODE and VTK

### LINKED ISSUE #673: Make VTK, CVODE and Xerces required dependencies (remove opt-out support)
Follow-up to #578 (VTK/CVODE) and #580 (Xerces). Consensus was to stop treating these as optional: every recommended install path already makes them easily available, and the Xerces opt-out doesn't even work today.

## What needs doing

- CMake: remove the `Chaste_USE_VTK`, `Chaste_USE_CVODE` and `Chaste_USE_XERCES` options from `CMakeLists.txt` and `cmake/Config/ChasteConfig.cmake.in`, making `find_package` and the corresponding `add_definitions` unconditional.
- Code: remove all `#ifdef CHASTE_VTK` / `CHASTE_CVODE` / `CHASTE_XERCES` guards and their no-op/stub branches. This is the bulk of the work, VTK and CVODE guards are spread across most modules; Xerces barely appears outside the CMake files.
- CI: drop the `01-no-cvode-vtk` portability matrix variant in `tier-1-portability.yml`. Check why `regenerate-archive.yml` explicitly disables CVODE before touching it.
- Docs: add a release notes entry for the breaking change.

## References

- #578: original VTK/CVODE discussion
- #580: Chaste_USE_XERCES=OFF is non-functional

fcooper8472:
Closes #673 

VTK, CVODE (SUNDIALS) and Xerces are now required dependencies, and the optional-dependency machinery is removed.

### Build system
- `Chaste_USE_VTK`, `Chaste_USE_CVODE` and `Chaste_USE_XERCES` options removed; the packages are always found. Setting any of them now prints a warning and is ignored (new `ChasteDeprecatedOptions.cmake`).
- `-DCHASTE_VTK`, `-DCHASTE_CVODE` and `-DCHASTE_XERCES` are still defined via a new `ChasteLegacyDefinitions.cmake`, deliberately kept for downstream projects. `TestChasteBuildInfo` has a new test that checks they stay defined.
- Minimum versions are now enforced at configure time: VTK 7.1, SUNDIALS 5.8, Xerces-C 3.2, XSD 4.0.0.
- Removed the old pychaste branch for VTK < 7, the unused XSD 3.2.0 search paths, and the stale Doxygen `PREDEFINED` entries.

### Source and tests
- Removed all `#ifdef CHASTE_VTK / CHASTE_CVODE / CHASTE_XERCES` guards, along with the no-op and fallback code they protected, across every module and their tests. This includes the no-VTK lung stub classes and the tests for them, and the EXPORT_CCM_CVODE indirection.
- Cell-cycle and SRN models now always use `CvodeAdaptor`.
- Simplified the version conditionals that the new minimums make redundant (VTK, SUNDIALS, Xerces/XSD), and removed `ChasteXsdVersion.hpp` and its test.
- Build info output: Xerces, SUNDIALS and VTK are now listed under `CompiledIn` (alphabetical, PT-Scotch after ParMETIS), and `Optional` is removed.
- Cleaned up stale comments, tutorial text and leftover formatting.

### CI and archive
- Removed the `no-cvode-vtk` portability variant and the CVODE-off flag from the archive-regeneration workflow.
- Regenerated the crypt archive. It changes a lot because the old one was generated with CVODE off (RK4 solver) and the new one with CVODE on. The generator test is chaotic, so this is expected. The final cell count is still within the test's asserted bounds.


fcooper8472:
@mirams and @jmpf: there's a reference left to `-DCHASTE_VTK`, in `ChasteParameters*.xsd`:

```xml
        <xs:documentation>Convert mesh and voltage data to parallel VTK format suitable for Paraview, Mayavi2
          etc (.pvtu file format). Requires that code has been compiled against vtkIO library (-DCHASTE_VTK).
          www.vtk.org</xs:documentation>
```

It's just in a docstring, so it doesn't really matter at all, but it would definitely be slightly neater if we tidied this up. That said, this isn't a file that changes often, and I don't know how to do that. A steer from one or both of you would be helpful!

jmpf:
> @mirams and @jmpf: there's a reference left to `-DCHASTE_VTK`, in `ChasteParameters*.xsd`:
> 
> ```
>         <xs:documentation>Convert mesh and voltage data to parallel VTK format suitable for Paraview, Mayavi2
>           etc (.pvtu file format). Requires that code has been compiled against vtkIO library (-DCHASTE_VTK).
>           www.vtk.org</xs:documentation>
> ```
> 
> It's just in a docstring, so it doesn't really matter at all, but it would definitely be slightly neater if we tidied this up. That said, this isn't a file that changes often, and I don't know how to do that. A steer from one or both of you would be helpful!

Bumping the XSD was a pain.  Maybe we should do it again, to learn what's involved.  However, it is probably best if we just remove the (-DCHASTE_VTK) from the string.  The rest of the string is still true.

fcooper8472:
> However, it is probably best if we just remove the (-DCHASTE_VTK) from the string. The rest of the string is still true.

@jmpf can I just edit the most recent xml file to just remove that word? or do I need to create a new one with a new version number etc?

fcooper8472:
@kwabenantim I had to make this change 82d05e300b0823eed1d8acef2736f45320bf3568 to pychaste.

Making VTK, CVODE, and Xerces required dependencies meant removing the `#ifdef CHASTE_CVODE` guards around CVODE-only code, including version-specific compatibility shims inside `CvodeAdaptor.hpp`.

`castxml` never received the same preprocessor definitions as a real build, so it always preprocessed those headers as if CVODE was unavailable. This was harmless before, because the removed guards meant the whole file compiled to nothing for that tool either way.

Once the guards were removed, the wrapper-generation tool started parsing the real CVODE code for the first time, including a shim that depends on knowing the installed SUNDIALS version. Without that information, the shim never activated, and parsing failed on newer SUNDIALS (as now shipped on Ubuntu 26.04) where an old type name no longer exists.

The fix makes the wrapper-generation tool see the same preprocessor definitions as the real build, so it parses the headers consistently with how they actually compile.

mirams:
Might find we need changes to `chaste_codegen` too (@kwabenantim can help with that).

jmpf:
> > However, it is probably best if we just remove the (-DCHASTE_VTK) from the string. The rest of the string is still true.
> 
> @jmpf can I just edit the most recent xml file to just remove that word? or do I need to create a new one with a new version number etc?

Yes.  I believe it's 
heart/src/io/ChasteParameters_2017_1.xsd

kwabenantim:
>  I had to make this change 82d05e3

That looks good to me, thanks @fcooper8472

kwabenantim:
> Might find we need changes to `chaste_codegen` too

Thanks. It appears to be passing, but I've opened an issue to look into whether anything more is required: https://github.com/ModellingWebLab/chaste-codegen/issues/309

mirams (approved):

